PR #59, opened 2023-09-12
Feature/cleaner network status logic

johnqh:
Cleaner logic with a single entry point function. It increments failure count when indexer or validator fails. When failure count > 3, status changes.


src/commonMain/kotlin/exchange.dydx.abacus/state/app/AppStateMachineProtocol.kt
```diff
@@ -216,44 +216,50 @@ enum class NetworkStatus(val rawValue: String) {
...
-            updateStatus()
+        } else {
+            failCount += 1
         }
+        updateStatus()
+    }
 
     private fun updateStatus() {
-        val requestTime = requestTime
         val time = time
```
ruixhuang, line 253: This can be deleted.

ruixhuang (approved):

PR #67, opened 2023-09-14
Fee tier volume should be in Quantums

rosepuppy:


src/commonMain/kotlin/exchange.dydx.abacus/processor/configs/FeeTierProcessor.kt
```diff
@@ -30,10 +30,10 @@ internal class FeeTierProcessor(parser: ParserProtocol) : BaseProcessor(parser)
         payload: IMap<String, Any>
     ): IMap<String, Any>? {
         val received = transform(existing, payload, feeTierKeyMap)
-        var absoluteVolumeRequirement = parser.asDecimal(payload["absoluteVolumeRequirement"])
+        val absoluteVolumeRequirement = parser.asDecimal(payload["absoluteVolumeRequirement"])
         if (absoluteVolumeRequirement != null) {
-            received["volume"] = absoluteVolumeRequirement
             received["symbol"] = "≥"
+            received["volume"] = absoluteVolumeRequirement / 1000000
```
jaredvu, line 36: Should we add a constant variable for `1_000_000` and use that for division in all of these cases. Would be more clear for people wondering why we are dividing each of these items by `1_000_000`

jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/processor/configs/FeeTierProcessor.kt
rosepuppy: yes, moved it to util.constants

PR #70, opened 2023-09-15
Kotlin toolchain update

ruixhuang:
Otherwise it won't work with Xcode 15

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -22,9 +22,9 @@ allprojects {
 }
 
 plugins {
-    kotlin("multiplatform") version "1.8.22"
-    kotlin("native.cocoapods") version "1.8.22"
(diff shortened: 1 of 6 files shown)


ruixhuang:
> Tested. A couple of things.
> 
> 1. We as well update gradle to 1.9.10 to be in sync with Kotlin.
> 2. Issue with earlier version of Kotlin 1.9 was the abacusjs.d.ts (Typescript stub) was not generated. This seems to be with 1.9.10. To test, run './gradlew assembleJsPackage' and look in /build/js/packages/abacusjs/abacusjs.d.ts
> 3. When you run assembleJsPackage, you will see an error. You need to run './gradlew kotlinUpgradeYarnLock' to generate the yarn lock file again.

OK.. Upudated... Also add steps to run /gradlew assembleJsPackage on CD/CI.

PR #72, opened 2023-09-18
TRCL-2620 emit error events

johnqh:
@jaredvu @mikemag-dev
FE should handle errorsEmitted function in StateNotificationProtocol interface. If it sees error.type as ParsingErrorType.HttpError403, the UX should handle according to requirements

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,7 +30,7 @@ plugins {
 }
 
 group = "exchange.dydx.abacus"
-version = "0.4.33"
+version = "0.4.44"
 
 repositories {
     google()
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -995,6 +1002,39 @@ class V4StateManagerAdaptor(
...
+
+                        emitError(error)
+                        trackApiCall()
+                    }
+                }
+
+                429 -> {
+                    /* We may need better handling of 429 */
+                    callback(response, httpCode)
+                }
```
jaredvu, line 1030: What is the rationale for not emitting an error for 429?

jaredvu (approved):

PR #87, opened 2023-09-26
Feature/trcl 2763 optimization

johnqh:
Use Double for everything except for raw data from the protocol.
Parser.asDouble() will handle super long string which would normally overflow double. (Have happened with indexPrice in v3)


src/commonTest/kotlin/exchange.dydx.abacus/utils/Map+Satisfy.kt
```diff
@@ -84,28 +85,38 @@ private fun IMap<String, Any>.satisfiesLists(
...
     path: String
 ) {
-    parser.asDouble(value1)?.let {
-        assertEquals(parser.asDouble(value2), it, "$path should have a value of $it")
-        return
+    parser.asDouble(value2)?.let {
+        val value = parser.asDouble(value1)
+        if (value != null) {
+            val stepSize = parser.asDouble(it.tickDecimals())!!
+            val roundedValue = Rounder.round(parser.asDouble(value2)!!, stepSize)
```
ruixhuang, line 96: We check null for value1.  Should we check null for value2 instead of "!!" ?

rosepuppy:
Looks like the version needs to be bumped again

src/commonMain/kotlin/exchange.dydx.abacus/utils/Parser.kt
```diff
@@ -86,11 +86,16 @@ class Parser : ParserProtocol {
 
         val string = data as? String
         if (string != null) {
+            val withDecimalPoint = string.replace(",", ".")
```
rosepuppy, line 89: What is the reason for this this?

rosepuppy (approved):
Rui should approve as well, but LGTM

ruixhuang (approved):

PR #95, opened 2023-10-06
Feature/trcl 2848 call screen every hour

johnqh:
Latest change according to Restriction project - calling screen every hour after NO_RESTRICTION, and also handles 403 error based on payload

Tested by integrating with iOS app (with local Abacus) and set the duration to 1 second. Saw the /screen endpoint being called every second

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
@@ -298,7 +298,7 @@ open class StateManagerAdaptor(
         }
 
     private val addressRetryDuration = 10.0
-    private val addressContinousMonitoringDuration = 60.0 * 60.0
+    private val addressContinuousMonitoringDuration = 60.0 * 60.0
 
     private var sourceAddressTimer: LocalTimerProtocol? = null
         set(value) {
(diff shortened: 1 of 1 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -296,6 +297,9 @@ open class StateManagerAdaptor(
             }
         }
 
+    private val addressRetryDuration = 10.0
+    private val addressContinousMonitoringDuration = 60.0 * 60.0
```
jaredvu, line 301:
```suggestion
    private val addressContinuousMonitoringDuration = 60.0 * 60.0
```

jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1063,24 +1066,39 @@ class V4StateManagerAdaptor(
...
+                        else
+                            UsageRestriction.http403Restriction
+                    } else UsageRestriction.http403Restriction
                 }
 
                 429 -> {
-                    /* We may need better handling of 429 */
-                    callback(response, httpCode)
+                    // retry after 5 seconds
+                    val timer = ioImplementations.timer ?: CoroutineTimer.instance
```
ruixhuang, line 1091: Optimization:  Might want to skip the timer run if there is already one pending... i.e.,

`if (restRetryTimers[url] == null) { ... }`

PR #97, opened 2023-10-09
Feature/trcl 2900 initializing state

johnqh:
Status mapping
ACTIVE/ONLINE -> canTrade: true, canReduce: true
CLOSE_ONLY -> canTrade:false, canReduce:true
INITIALIZING -> canTrade:false, canReduce:false

Abacus returns markets with either canTrade or canReduce. FE app should filter to display canTrade only in the list. Navigation is OK for market with either canTrade or canReduce
AsyncStateManager now will not allow setting market field if neither canTrade nor canReduce.

@jaredvu @ruixhuang Previously, the market list returns canTrade markets only, now it returns either canTrade or canReduce. FE needs to be changed to do one more filtering, to canTrade only

### DIFF
--- a/src/commonTest/kotlin/exchange.dydx.abacus/payload/BaseTests.kt
+++ b/src/commonTest/kotlin/exchange.dydx.abacus/payload/BaseTests.kt
@@ -246,7 +246,11 @@ open class BaseTests(private val maxSubaccountNumber: Int) {
             assertEquals(parser.asString(data["current"]), obj.current?.rawValue, "$trace.current")
             when (obj.current?.rawValue) {
                 "trade" -> {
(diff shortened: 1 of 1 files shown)


jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/AsyncAbacusStateManager.kt
```diff
@@ -637,6 +639,15 @@ class AsyncAbacusStateManager(
         adaptor?.transfer(data, type)
     }
 
+    fun isMarketValid(marketId: String?): Boolean {
```
ruixhuang, line 642: private?

PR #113, opened 2023-10-17
Feature/trcl 2432 localized string

johnqh:


src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -429,7 +447,15 @@ data class SubaccountOrderResources(
                         existing.statusStringKey != statusStringKey ||
                         existing.timeInForceStringKey != timeInForceStringKey
                     ) {
+                        val sideString = localizer?.localize(sideStringKey)
+                        val typeString = if (typeStringKey != null) localizer?.localize(typeStringKey) else null
```
ruixhuang, line 451: Nit: "if" check is not needed.

ruixhuang (approved):

PR #114, opened 2023-10-17
Feature/trcl 3016 denom in connect

johnqh:
Use denom settings from configs.

1. DYDXChainTransactionsProtocol connectNetwork interface changed, and now uses paramsInJson, so we can add more params in the future without breaking integration. This requires FE app to modify integrations.
2. There was a recursion crash with iOS integration, due to environmentsChanged getting called in constructor. Changed so environmentsChanged is only called after initialization (fetched from remote sources). When AsyncAbacusManager is initialized, it should already have a list of environments from app bundle. (iOS) when it fetches a new list from remote, environmentsChanged gets called.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
@@ -498,10 +498,10 @@ class V4StateManagerAdaptor(
         val usdcToken = environment.tokens["usdc"] ?: return
         val chainToken = environment.tokens["chain"] ?: return
         val usdcDenom = usdcToken.denom
-        val usdcDenomExponent = usdcToken.exponents
+        val usdcDecimals = usdcToken.decimals
         val usdcGasDenom = usdcToken.gasDenom
         val chainTokenDenom = chainToken.denom
-        val chainTokenDenomExponent = chainToken.exponents
+        val chainTokenDecimals = chainToken.decimals
 
         val params = mutableMapOf<String, Any>()
         params["indexerUrl"] = indexerUrl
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/AsyncAbacusStateManager.kt
```diff
@@ -193,13 +193,18 @@ class AsyncAbacusStateManager(
...
 
+    private var initializing: Boolean = true
     private var environments: IList<V4Environment> = iListOf()
         set(value) {
             field = value
-            stateNotification?.environmentsChanged()
-            dataNotification?.environmentsChanged()
             ioImplementations.threading?.async(ThreadingType.abacus) {
                 _environment = findEnvironment(environmentId)
+                if (!initializing) {
```
ruixhuang, line 202: iOS app is already switching to main thread at environmentsChanged().

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -493,13 +495,31 @@ class V4StateManagerAdaptor(
         val websocketUrl = configs.websocketUrl() ?: return
         val chainId = environment.dydxChainId ?: return
         val faucetUrl = configs.faucetUrl()
+        val usdcToken = environment.tokens["usdc"] ?: return
+        val chainToken = environment.tokens["chain"] ?: return
+        val usdcDenom = usdcToken.denom
+        val usdcDenomExponent = 6
+        val usdcGasDenom = usdcToken.gasDenom
+        val chainTokenDenom = chainToken.denom
+        val chainTokenDenomExponent = 18
```
ruixhuang, line 504: Are we getting this from configs?

ruixhuang (approved):

PR #115, opened 2023-10-17
Feature/trcl 3013 handling transfers in abacus

johnqh:
1. Account balances are now returning string instead of Double.
2. Transfer/Deposit/Withdrawal amount are now string instead of double

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/modal/TradingStateMachine.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/modal/TradingStateMachine.kt
@@ -541,7 +541,7 @@ open class TradingStateMachine(
         val input = state?.input
 
 
-        state = update(state, changes, environment?.tokens?.get("chain")!!, localizer)
+        state = update(state, changes, chainTokenInfo, localizer)
 
         val realChanges = iMutableListOf<Changes>()
         for (change in changes.changes) {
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/modal/TradingStateMachine.kt
```diff
@@ -536,7 +540,8 @@ open class TradingStateMachine(
         val wallet = state?.wallet
         val input = state?.input
 
-        state = update(state, changes, localizer)
+
+        state = update(state, changes, environment?.tokens?.get("chain")!!, localizer)
```
ruixhuang, line 544: Nit: Replace with `chainTokenInfo`

src/commonMain/kotlin/exchange.dydx.abacus/utils/Parser.kt
```diff
@@ -107,38 +112,43 @@ class Parser : ParserProtocol {
         if (decimal != null) {
             return decimal
         }
+
+        val string = data as? String
+        if (string != null) {
+            return try {
+                string.toBigDecimal(null, null)
+            } catch (e: Exception) {
+                DebugLogger.error("Failed to parse double: $string", e)
```
ruixhuang, line 121: Failed to parse "decimal"

ruixhuang (approved):

PR #116, opened 2023-10-18
Feature/trcl 3022 duplicate trigger and index price validation

johnqh:


src/commonMain/kotlin/exchange.dydx.abacus/validator/trade/TradeTriggerPriceValidator.kt
```diff
@@ -227,15 +227,16 @@ internal class TradeTriggerPriceValidator(
         )
     }
 
-    private fun triggerToLimitError(
-        triggerToLimit: RelativeToPrice,
-        triggerLimit: Double,
+    private fun triggerToLiquidationError(
```
jaredvu, line 230: 🔥

jaredvu (approved):

PR #120, opened 2023-10-20
Fix findTokensInfo

rosepuppy:


src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -1443,7 +1443,7 @@ data class Account(
         }
 
         private fun findTokenInfo(tokensInfo: Map<String, TokenInfo>, denom: String): TokenInfo? {
-            return tokensInfo.values.firstOrNull() { it.denom == denom }
```
rosepuppy, line 1446: tokensInfo.values was returning empty list for some reason when I was testing integration on web

PR #126, opened 2023-10-25
Feature/trcl 3101 filter market without configs

johnqh:


src/commonTest/kotlin/exchange.dydx.abacus/payload/BaseTests.kt
```diff
@@ -1266,6 +1276,10 @@ open class BaseTests(private val maxSubaccountNumber: Int) {
                 "$trace.perpetual"
             )
         } else {
+            if (obj != null) {
+                val x = 0
```
ruixhuang, line 1280: Delete

ruixhuang (approved):

PR #130, opened 2023-10-27
TRCL-3120 Add Squid V2 asset fetching

ruixhuang:
Squid version can be toggled by the client using AppConfigs.  Will add a feature flag on iOS.

Right now, the V2 URLs are hardcoded.  Chain and token fetching work, but route fetching doesn't work with dydx chain.


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor+Transfer.kt
```diff
@@ -41,10 +45,11 @@ internal fun V4StateManagerAdaptor.retrieveDepositRoute(state: PerpetualState?)
         fromAmountString != null &&
         accountAddress != null &&
         chainId != null &&
+        dydxTokenDemon != null &&
```
rosepuppy, line 48: curious, what is the motivation for this check?

rosepuppy (approved):
awesome!

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor+Transfer.kt
ruixhuang: The token denom comes from the env json. All fields are optional for json parsing

PR #131, opened 2023-10-27
TRCL-3113 Run a timer to refresh historical funding at hour

johnqh:


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1318,9 +1332,55 @@ open class StateManagerAdaptor(
             if (success(httpCode) && response != null) {
                 update(stateMachine.historicalFundings(response), oldState)
             }
+            ioImplementations.threading?.async(ThreadingType.main) {
```
ruixhuang, line 1335: Can this get called multiple times?  If so, might want to set historicalFundingTimer to null before starting another timer.

ruixhuang (approved):

PR #151, opened 2023-11-10
Feature/trcl 3209 order type changes

johnqh:


src/commonMain/kotlin/exchange.dydx.abacus/output/Market.kt
```diff
@@ -550,7 +551,7 @@ data class MarketTrade(
     val side: OrderSide,
     val size: Double,
     val price: Double,
-    val liquidation: Boolean,
```
ruixhuang, line 553: This would make it a breaking change, right?  Should we update the major version?

build.gradle.kts
```diff
@@ -30,7 +30,7 @@ plugins {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.0.25"
+version = "1.0.26"
```
jaredvu, line 33:
```suggestion
version = "1.1.0"
```

jaredvu (approved):

PR #155, opened 2023-11-22
[TRCL-3254] Features/squid transaction status

ruixhuang:
Going to use it on the client to check the status in addition to the check of the last route step.


rosepuppy (approved):
needs rebase, LGTM otherwise

PR #157, opened 2023-11-28
TRCL-3240 Toggle between Squid V1 and V2 Routes

ruixhuang:
The client should set the appConfigs.squidVersion (V2, V2DepositOnly or V2WithdrawOnly).  By default, Abacus uses V1 to retain the current behavior. @rosepuppy 

CCTP withdrawal route fetch is returning error, but deposit route works.


rosepuppy (approved):
LGTM!

PR #160, opened 2023-11-30
Poll and sweep noble balance

rosepuppy:


src/commonMain/kotlin/exchange.dydx.abacus/output/input/TransferInput.kt
```diff
@@ -571,6 +571,7 @@ data class TransferInput(
     val summary: TransferInputSummary?,
     val resources: TransferInputResources?,
     val requestPayload: TransferInputRequestPayload?,
+    val cctp: Boolean,
```
ruixhuang, line 574: This won't be needed.  There is already an extension property that does this:

`val TransferInput.isCctp: Boolean`

It looks at the TransferInput's token and chain to determine if the route is CCTP.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -409,6 +420,30 @@ class V4StateManagerAdaptor(
         }
     }
 
+    private fun pollNobleBalance() {
```
ruixhuang, line 423: Nit: Move those functions to the V4StateManagerAdaptor+Transfer.kt.

PR #165, opened 2023-12-04
TRCL-3356 Do not update apiStatus if app might have just been in background.

johnqh:
It has been a common complaint that the app shows an INDEX_TAILING notification when it comes from background. On web, it means it was in a background tab, or browser was minimized. On iOS, it means it was in background.

I have added logic to detect whether the app has been in background prior to the current network status update, and do not update to INDEX_TAILING in this case.

Tested with iOS integration. Web dev please do some testing with web app integration.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,7 +30,7 @@ plugins {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.1.10"
+version = "1.1.11"
 
 repositories {
     google()
--- a/v4_abacus.podspec
+++ b/v4_abacus.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'v4_abacus'
-    spec.version                  = '1.1.10'
+    spec.version                  = '1.1.11'
     spec.homepage                 = 'https://github.com/dydxprotocol/v4-abacus'
     spec.source                   = { :http=> ''}
     spec.authors                  = ''


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -717,6 +727,35 @@ class V4StateManagerAdaptor(
         }
     }
 
+    private fun delayedInRequestTime(networkState: NetworkState): Boolean {
```
adamfraser, line 730: Feels like a bit of a hacky way to do this, using random numbers for the time duration. Can't we use the block numbers being returned from indexer and validator endpoints to determine this?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -717,6 +727,35 @@ class V4StateManagerAdaptor(
...
+        val now = Clock.System.now()
+        val time = networkState.requestTime
+        return if (time != null) {
+            val gap = now - time
+            if (gap > 4.seconds) {
+                // If request (time) was sent more than 4 seconds ago
+                // The app was probably in background
+                true
+            } else {
+                val previousTime = networkState.prevousRequestTime
```
ruixhuang, line 740: Nit: previousRequestTime

ruixhuang (approved):

PR #174, opened 2023-12-08
Handle CCTP withdraw

rosepuppy:
- expose a method for starting a cctp withdraw
- generate squid route to noble and send the transaction
- set a flag that repurposes pollNobleBalance to instead send the squid withdraw transaction
- `commitCCTPWithdraw` callback returns the transaction hash needed for squid status


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor+Transfer.kt
```diff
@@ -453,3 +456,93 @@ internal fun V4StateManagerAdaptor.transferNobleBalance(amount: BigDecimal) {
...
+internal fun clearCctpWithdraw() {
+    isCctpWithdraw = false
+    cctpWithdrawPayload = null
+    cctpWithdrawCallback = null
+}
+internal fun V4StateManagerAdaptor.cctpToNoble(
+    state: PerpetualState?,
+    decimals: Int,
+    gas: BigDecimal,
+    callback: TransactionCallback
```
ruixhuang, line 473: Since caller is supplying the callback, can we make sure the callback is invoked on every code path?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor+Transfer.kt
```diff
@@ -453,3 +456,93 @@ internal fun V4StateManagerAdaptor.transferNobleBalance(amount: BigDecimal) {
...
+                            "amount" to state?.input?.transfer?.size?.usdcSize,
+                            "ibcPayload" to ibcPayload,
+                        )
+                    )
+                    transaction(TransactionType.WithdrawToNobleIBC, payload) {
+                        val error = parseTransactionResponse(it)
+                        if (error != null) {
+                            DebugLogger.error("transferNobleBalance error: $error")
+                            send(error, callback)
+                            clearCctpWithdraw()
```
ruixhuang, line 533: Nit: Instead of clearing it on error... we could store the pending state only when `transaction(TransactionType.WithdrawToNobleIBC` succeeds.  I think it would make the code path easier construct.

ruixhuang (approved):

PR #183, opened 2023-12-13
Feature/trcl 3391 fill blank rewards

johnqh:
Indexer will skip the periods without any rewards. Abacus would fill in the blanks.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,7 +30,7 @@ plugins {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.1.25"
+version = "1.1.26"
 
 repositories {
     google()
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -1488,6 +1555,210 @@ data class TradingRewards(
...
+        }
+
+        private fun currentPeriodPlaceHolder(period: String): HistoricalTradingReward {
+            val now = Clock.System.now()
+            val thisPeriod = when (period) {
+                "DAILY" -> today(now)
+                "WEEKLY" -> thisWeek(now)
+                "MONTHLY" -> thisMonth(now)
+                else -> today(now)
+            }
```
aforaleka, line 1658: to clarify, shouldn't we always know the current period's start time from the api response? (i.e. the most recent period should be the current period) so we should only need to calculate the end time?

aforaleka (approved):

PR #202, opened 2024-01-22
needsReduceOnly logic and tests

johnqh:


src/commonMain/kotlin/exchange.dydx.abacus/output/input/TradeInput.kt
```diff
@@ -70,6 +69,7 @@ data class TradeInputOptions(
     val timeInForceOptions: IList<SelectionOption>?,
     val goodTilUnitOptions: IList<SelectionOption>,
     val executionOptions: IList<SelectionOption>?,
+    val reduceOnlyPromptStringKey: String?
```
mike-dydx, line 72: @johnqh how come we have this for `reduceOnly` but not `postOnly`?

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
```diff
@@ -1027,6 +1039,11 @@ internal class TradeInputCalculator(
             } else {
                 options.safeSet("maxLeverage", null)
             }
+            if (parser.asBool(options["needsReduceOnly"]) == true) {
+                options.safeSet("reduceOnlyPromptStringKey", null)
+            } else {
+                options.safeSet("reduceOnlyPromptStringKey", reduceOnlyPromptFromTrade(trade))
+            }
```
mike-dydx, line 1046: is this reversed? if `needsReduceOnly` is true, I'm thinking  `reduceOnlyPromptStringKey` should be non-nil so client knows what to display

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
```diff
@@ -827,32 +836,32 @@ internal class TradeInputCalculator(
...
-                trade,
-                "execution"
-            )
-        )
-        return when (execution) {
-            "FOK", "IOC" -> reduceOnlyField()
-            else -> null
-        }
-    }
+//
```
mike-dydx, line 839: delete or intentionally commented?

jaredvu (approved):
Which ver of v4-localization do we need with this ver of v4-abacus?

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
mike-dydx: ah i see. Small suggestion then to rename to something like `reduceOnlyUnavailablePromptStringKey` but not strong feelings on that

mike-dydx (approved):

PR #209, opened 2024-01-26
Feature/execution default

johnqh:
Previously, the execution default was set at wrong place, so it was not set when going from STOP_LIMIT/TAKE_PROFIT to STOP_MARKET/TAKE_PROFIT_MARKET.

Logic change + extensive unit tests to cover all cases


src/commonTest/kotlin/exchange.dydx.abacus/payload/v4/V4TradeInputTests.kt
```diff
@@ -1174,4 +1175,804 @@ open class V4TradeInputTests : V4BaseTests() {
         )
 
     }
+
+
+    private fun testExecution() {
+        testExecutionStopLimit()
```
ruixhuang, line 1181: Are there dependencies between those tests?  If not, can we make them public instead of private, so we can run them separately?

ruixhuang (approved):

PR #219, opened 2024-02-12
Feature/trcl 2044 default 100percent close position

johnqh:
closePosition set to initial 100%

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,7 +30,7 @@ plugins {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.4.8"
+version = "1.4.9"
 
 repositories {
     google()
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/changes/StateChanges.kt
```diff
@@ -48,4 +49,19 @@ data class StateChanges(
     companion object {
         val noChange = StateChanges(iListOf<Changes>())
     }
+
+    fun merge(earlierChanges: StateChanges): StateChanges {
```
aforaleka, line 53: what's the context for this function?

aforaleka (approved):

PR #224, opened 2024-02-20
Feature/trcl 3552 get launch incentive seasons

johnqh:
@aforaleka Seaons is a root object under State. Points are under Account

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/launchIncentive/LaunchIncentiveProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/launchIncentive/LaunchIncentiveProcessor.kt
@@ -1,0 +1,49 @@
+package exchange.dydx.abacus.processor.launchIncentive
+
+import exchange.dydx.abacus.processor.base.BaseProcessor
+import exchange.dydx.abacus.processor.wallet.account.LaunchIncentivePointsProcessor
+import exchange.dydx.abacus.protocols.ParserProtocol
+import exchange.dydx.abacus.utils.mutable
+
+internal class LaunchIncentiveProcessor(parser: ParserProtocol) : BaseProcessor(parser) {
+    private val seasonsProcessor = LaunchIncentiveSeasonsProcessor(parser = parser)
+    private val pointsProcessor = LaunchIncentivePointsProcessor(parser = parser)
+
+    internal fun receivedSeasons(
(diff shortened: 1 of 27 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/processor/launchIncentive/LaunchIncentiveProcessor.kt
```diff
@@ -0,0 +1,49 @@
...
+                 },
+                 {
+                    "startTimestamp":1704384000,
+                    "label":"2",
+                    "__typename":"TradingSeason"
+                 }
+              ]
+           }
+        }
+         */
```
mike-dydx, line 34: since this is in the test data file, should we remove? Unless you think it is helpful to see in-line with the processor code

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1390,6 +1394,40 @@ class V4StateManagerAdaptor(
         }
     }
 
+    private fun retrieveLaunchIncentiveSeasons() {
+        val url = configs.launchIncentiveUrl("graphql")
+        if (url != null) {
+            val requestBody = "{\"operationName\":\"TradingSeasons\",\"variables\":{},\"query\":\"query TradingSeasons {tradingSeasons {startTimestamp label __typename }}\"}"
```
mike-dydx, line 1400: nit: can are you able remove `__typename` from the raw graphql query? or does that break the query

src/commonTest/kotlin/exchange.dydx.abacus/payload/BaseTests.kt
```diff
@@ -192,9 +193,9 @@ open class BaseTests(private val maxSubaccountNumber: Int) {
         }
     }
 
-    internal fun verifyState(state: PerpetualState?) {
+    internal open fun verifyState(state: PerpetualState?) {
         verifyConfigs(perp.configs, state?.configs, "configs")
-        verifyWalletState(perp.wallet, state?.wallet, "wallet")
+//        verifyWalletState(perp.wallet, state?.wallet, "wallet")
```
mike-dydx, line 198: intentional?

mike-dydx (approved):

mike-dydx (approved):

PR #232, opened 2024-03-04
MOB-301 : update `connectNetwork` parameters to pass in memo field

mike-dydx:
[MOB-301 : update `connectNetwork` parameters to pass in memo field](https://linear.app/dydx/issue/MOB-301/update-connectnetwork-parameters-to-pass-in-memo-field)

Screenshot 2024-03-04 at 6 51 45 PM


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/SystemUtils.kt
```diff
@@ -0,0 +1,12 @@
+package exchange.dydx.abacus.state.manager
+
+enum class Platform(val rawValue: String) {
+    ios("ios"),
+    android("android"),
+    web("web"),
+}
+expect class SystemUtils {
```
mike-dydx, line 8: `expect` here is a way for Kotlin Multiplat to define a protocol which expects an `actual` implementation for each supported platform

src/iosMain/kotlin/exchange.dydx.abacus/state/manager/SystemUtils.ios.kt
```diff
@@ -0,0 +1,7 @@
+package exchange.dydx.abacus.state.manager
+
+actual class SystemUtils {
```
mike-dydx, line 3: `actual` here is the "actualy" implementation of the `expect` protocol

src/nativeMain/AndroidManifest.xml
```diff
@@ -1,2 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<manifest package="exchange.dydx.abacus.android" />
```
mike-dydx, line 2: this was moved to `androidMain` instead of `nativeMain`

ruixhuang (approved):
👍

PR #235, opened 2024-03-11
Feature/mob 323 unit tests for network requests

johnqh:
This should be a safe PR. It does three things

1. Move the structures from AsyncAbacusManager to separate files
2. Created an interface for AsyncAbacusManager, so I can create a new implementation for the interface without touching the existing one.
3. Have better and more complete unit testing to verify network requests.

Also there are some manual linting. Auto-linting will come in separate PR.


src/commonTest/kotlin/exchange.dydx.abacus/app/manager/TestFactory.kt
```diff
@@ -341,6 +343,7 @@ class TestChain : DYDXChainTransactionsProtocol {
         paramsInJson: String?,
         callback: (response: String?) -> Unit
     ) {
+        requests.add(type)
```
mikemag-dev, line 346: is this getting used anywhere?

mikemag-dev (approved):

ruixhuang (approved):

PR #243, opened 2024-03-13
Feature/mob 246 refactoring subaccount handling

johnqh:
Tested `AsyncAbacusStateManagerV2` with iOS app integration, and ensured that subaccount functionalities work, including faucet, state calculations, and trade.

Noticeable changes:
`SubaccountSupervisor` has a `realized` flag, indicating whether there is subaccount payload from indexer. I take this approach because we need it to handle faucet before subaccount is created.

Fixed an exception bug with parseTransfersToMatchFaucetRecords, in both older code `AsyncAbacusStateManager` and new `AsyncAbacusStateManagerV2`. This exception was caught inside Abacus try block so it didn't cause crashes, but it's better not to throw exceptions.

Not yet done:

Test integration with FE for deposit and withdraw flow. Until then, FE apps should still use `AsyncAbacusStateManager`

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
@@ -303,9 +303,7 @@ internal open class AccountSupervisor(
 
         if (subaccounts.isNotEmpty()) {
             for ((subaccountNumber, subaccount) in subaccounts) {
-                if (canConnectTo(subaccountNumber)) {
-                    subaccount.realized = true
-                }
+                subaccount.realized = canConnectTo(subaccountNumber)
             }
         } else {
             subaccountNumber = 0


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -843,6 +862,11 @@ internal class SubaccountSupervisor(
                 }
             }
             update(changes, oldState)
+
+            val lastOrderClientId = this.lastOrderClientId
+            if (lastOrderClientId != null) {
```
jaredvu, line 867: Can we use `this.lastOrderClientId != null` instead of initializing a `val`?

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -292,8 +301,15 @@ internal open class AccountSupervisor(
         val oldState = stateMachine.state
         update(stateMachine.account(response), oldState)
 
-        // Automatically connect to the first subaccount if no subaccount is connected
-        updateConnectedSubaccountNumber()
+        if (subaccounts.isNotEmpty()) {
+            for ((subaccountNumber, subaccount) in subaccounts) {
+                if (canConnectTo(subaccountNumber)) {
+                    subaccount.realized = true
```
jaredvu, line 307: Possible to simplify to `subaccount.realized = canConnectTo(subaccountNumber)`?

jaredvu (approved):

PR #248, opened 2024-03-18
Add stateful orders transaction queue to allow multiple orders cancel/placement at a time

aforaleka:
context: we're seeing account seq mismatch broadcast errors when stateful orders are placed or canceled in quick succession, so usually first order cancel goes through but subsequent order cancels fail. adding a tx queue can help mitigate this by submitting place/cancel order one at a time while not blocking user from submitting multiple tx.

- for stateful orders (i.e. not short term), transaction is added to `StatefulOrdersTransactionQueue`, and run sequentially
- adds additional amplitude tracking to differentiate place order click vs. submit order


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StatefulOrdersTransactionQueue.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    private fun processNext() {
+        if (queue.isEmpty()) {
+            isProcessing = false
+            return
+        }
+
+        val currentTransaction = queue.removeAt(0)
+
+        val submitTimeMs = Clock.System.now().toEpochMilliseconds().toDouble()
```
prashanDYDX, line 39: Why are we converting to Double here (and in the other usages)?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StatefulOrdersTransactionQueue.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlinx.datetime.Clock
+
+class TransactionParams(
+    val type: TransactionType,
+    val payload: String,
+    val callback: (response: String, uiDelayTimeMs: Double, submitTimeMs: Double) -> Unit,
+    val uiClickTimeMs: Double
+)
+
+class StatefulOrdersTransactionQueue(
```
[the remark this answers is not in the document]
aforaleka: ok! renaming to `TransactionQueue`, i think it'll probably just work as is, as long as that transaction only finishes after both msgs are sent

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -961,8 +962,9 @@ class V4StateManagerAdaptor(
         }
     }
 
+    open val transactionQueue = TransactionQueue(this::transaction)
```
prashanDYDX, line 965: Instead of making this open, we can pass it into the constructor.

PR #258, opened 2024-03-21
1.6.16: Add TriggerOrdersInput to abacus state machine

moo-onthelawn:
Add a skeleton TriggerOrdersInput to abacus. 
- Does not yet
    - Do any of the calculations 
    - Do any validation
    - Expose the commitTriggerOrders endpoint or payload functions
- Does
    - Fill out documentation 
    - Hook up/expose the `tradingStateMachine.triggerOrders` function to allow the front end to set and read values from `triggerOrdersInput`
    - set the `price.input` flag with the user's most recently updated `triggerPrice` field (can be the price, percent diff or price diff)
    - write a very basic input update test
 
Tested out by hooking up to a local version of v4-web (where I could read and update abacus values). I think* these are all the attributes we need (I don't think we need `summary` - we should be able to just update the existing properties when calculating the limit price for market orders).


gradle.properties
```diff
@@ -1,6 +1,4 @@
 kotlin.code.style=official
 kotlin.js.generate.executable.default=false
 org.gradle.jvmargs=-Xmx2000m
-kotlin.native.ignoreDisabledTargets=true
-org.gradle.caching=true
-org.gradle.parallel=true
```
moo-onthelawn, line 6: this happened when running the scripts; unsure if this change is an issue 🤔

aforaleka (approved):
lgtm! will need to bump an extra version i think

PR #257, opened 2024-03-21
Provide a groupedSubaccounts list, which transforms child subaccounts as isolated positions (after #255)

johnqh:
Review after #255

PR #265, opened 2024-03-27
Provides a list of SubaccountPendingPosition for child subaccounts with margin, pending order, but no position yet

johnqh:
After #257

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -47,7 +47,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.6.21"
+version = "1.6.22"
 
 repositories {
     google()
--- a/v4_abacus.podspec
+++ b/v4_abacus.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'v4_abacus'
-    spec.version                  = '1.6.21'
+    spec.version                  = '1.6.22'
     spec.homepage                 = 'https://github.com/dydxprotocol/v4-abacus'
     spec.source                   = { :http=> ''}
     spec.authors                  = ''


src/commonMain/kotlin/exchange.dydx.abacus/calculator/AccountCalculator.kt
```diff
@@ -19,17 +18,229 @@ class AccountCalculator(val parser: ParserProtocol) {
...
+                    val parentSubaccountNumber = subaccountNumber % NUM_PARENT_SUBACCOUNTS
+                    val parentSubaccount = parser.asNativeMap(
+                        parser.value(
+                            groupedSubaccounts,
+                            "$parentSubaccountNumber",
+                        ),
+                    ) ?: parser.asNativeMap(parser.value(subaccounts, "$parentSubaccountNumber"))
+                        ?: mapOf("subaccountNumber" to parentSubaccountNumber)
+
+                    val childOpenPositions =
```
jaredvu, line 70: `childOpenPositions` should only have a size of 1 unless they API traded. If they are using a `childSubaccount` as a `CROSS` margined acc, should we exclude from the merged positions in the Parent?

src/commonMain/kotlin/exchange.dydx.abacus/calculator/AccountCalculator.kt
```diff
@@ -19,17 +18,229 @@ class AccountCalculator(val parser: ParserProtocol) {
...
+            "equity",
+            sum(
+                parser.asMap(parentSubaccount["equity"]),
+                parser.asMap(childSubaccount["equity"]),
+            ),
+        )
+        return modifiedParentSubaccount
+    }
+
+    private fun sum(value1: Map<String, Any>?, value2: Map<String, Any>?): Map<String, Any> {
```
jaredvu, line 224: instead of `Map<String, Any>?` would it be possible to use `TradeStatesWithDoubleValues?`

jaredvu: I see got it!

src/commonMain/kotlin/exchange.dydx.abacus/calculator/AccountCalculator.kt
jaredvu: > The subaccount itself doesn't indicate whether the margin is isolated or cross. I don't think our UI as designed will cover all cases.

My suggestion was more for if we should ignore adding a `childSubaccount` to the `groupedAccounts` if we see that it is being used as `CROSS` margined acc. This may be difficult because we would have to ignore it for the other subaccount endpoints as well.

jaredvu (approved):

jaredvu (approved):

PR #267, opened 2024-03-27
v1.7.0: Placeorder flow for new margin mode (Merge into #256)

jaredvu:
* Add subaccount constants
*  Add `getChildSubaccountNumberForIsolatedMarginTrade` and `getTransactionsForIsolatedOrder`
  * Used within `placeOrderPayload` and `commitPlaceOrder`
* Update TransactionTest to ensure new methods return correct payload and transactions

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
@@ -50,6 +50,8 @@ import exchange.dydx.abacus.utils.GoodTil
 import exchange.dydx.abacus.utils.IList
 import exchange.dydx.abacus.utils.IMap
 import exchange.dydx.abacus.utils.IMutableList
+import exchange.dydx.abacus.utils.MAX_SUBACCOUNT_NUMBER
+import exchange.dydx.abacus.utils.NUM_PARENT_SUBACCOUNTS
 import exchange.dydx.abacus.utils.ParsingHelper
 import exchange.dydx.abacus.utils.iMapOf
 import exchange.dydx.abacus.utils.mutable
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -432,6 +434,82 @@ internal class SubaccountSupervisor(
...
+        val existingSubaccountNumbers = utilizedSubaccountsMarketIdMap?.keys ?: iListOf(subaccountNumber)
+        for (offset in NUM_PARENT_SUBACCOUNTS..MAX_SUBACCOUNT_NUMBER step NUM_PARENT_SUBACCOUNTS) {
+            val tentativeSubaccountNumber = offset + subaccountNumber
+            if (!existingSubaccountNumbers.contains(tentativeSubaccountNumber.toString())) {
+                return tentativeSubaccountNumber
+            }
+        }
+
+        // User has reached the maximum number of childSubaccounts for their current parentSubaccount
+        throw Exception("No available subaccount number")
```
prashanDYDX, line 488: kotlin tip: you can just user `error("No available subaccount number)`

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -432,6 +434,82 @@ internal class SubaccountSupervisor(
         }
     }
 
+    /**
+     * @description Get the childSubaccount number that is available for the given marketId
+     * @param marketId
+     */
+    @Throws(Exception::class)
```
prashanDYDX, line 441: `@Throws` here isn't super necessary - it's mostly for Java interop which we don't really care about. (As you can see from your call-site, Kotlin isn't forcing you to handle the exception)

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -432,6 +434,82 @@ internal class SubaccountSupervisor(
...
+        // User has reached the maximum number of childSubaccounts for their current parentSubaccount
+        throw Exception("No available subaccount number")
+    }
+
+    @Throws(Exception::class)
+    internal fun getTransactionsForIsolatedOrder(orderPayload: HumanReadablePlaceOrderPayload): IList<Transaction> {
+        val trade = stateMachine.state?.input?.trade
+
+        // Derive transfer params from trade input
+        val targetLeverage = trade?.targetLeverage ?: throw Exception("targetLeverage is null")
```
prashanDYDX, line 496: Same here

PR #268, opened 2024-03-27
Remove Napier and replace it with our own simple logging implementation.

prashanDYDX:
All the different logging levels seemed unnecessary, so only have two to start: debug and error. Can add more later if we need. The logger takes in a lambda so that we're not unnecessarily building strings if we don't need them.

Platform specific logging implementations are in Logger.jvm.kt, Logger.ios.kt, and Logger.js.kt. 

Feel free to make changes to these as you see fit.


prashanDYDX:
There's an API change as well (changing String param to () -> String), so we'll have the pain regardless.

src/commonMain/kotlin/exchange.dydx.abacus/utils/Logger.kt
```diff
@@ -0,0 +1,20 @@
+package exchange.dydx.abacus.utils
+
+object Logger {
+
+    var isDebugEnabled: Boolean = false
+
+    fun d(message: () -> String) {
+        if (isDebugEnabled) {
+            platformDebugLog(message())
```
ruixhuang, line 9: If we add LoggingProtocol to as an optional to IOImplementations, we can ask the client to implement their own logging.  We can still use platformDebugLog as a default if LoggingProtocol is not provided.

This would give client a lot more flexibility (e.g., on mobile we can log to files, or send event to Firebase)

prashanDYDX: I think I would want to separate logging from error reporting to reduce mental burden for developers. For example, don't want Web engineers to have to think about whether a log will get sent to Firebase on mobile or not. 

I think it'd be better to have a RemoteErrorReportingProtocol that we can call intentionally in places we care to send events. Wdyt?

ruixhuang (approved):

PR #269, opened 2024-03-28
skip retreive market candles based on appconfigs

aforaleka:
- didn't realize we have the `subscribeToCandles` field in V1, we should also skip calling the candles API on indexer config / market set for web
- add back the two properties that were previously removed -- i think that is related to build performance
- fixed yarn lock file, which prev. throws an error when building for js

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,3 +2,6 @@ kotlin.code.style=official
 kotlin.js.generate.executable.default=false
 org.gradle.jvmargs=-Xmx2000m
 kotlin.native.ignoreDisabledTargets=true
+
+org.gradle.caching=true
+org.gradle.parallel=true
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
@@ -421,7 +421,7 @@ open class StateManagerAdaptor(
             retrieveFeeTiers()
             if (market != null) {
                 retrieveMarketHistoricalFundings()
-                retrieveMarketCandles()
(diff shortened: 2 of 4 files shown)


gradle.properties
```diff
@@ -1,4 +1,7 @@
 kotlin.code.style=official
 kotlin.js.generate.executable.default=false
 org.gradle.jvmargs=-Xmx2000m
-kotlin.native.ignoreDisabledTargets=true
\ No newline at end of file
+kotlin.native.ignoreDisabledTargets=true
+
+org.gradle.caching=true
+org.gradle.parallel=true
```
moo-onthelawn, line 7: oh whoops I think this one was my bad - ty 🙏

moo-onthelawn (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1614,7 +1614,7 @@ open class StateManagerAdaptor(
                 }
             }
             retrieveMarketHistoricalFundings()
-            retrieveMarketCandles()
+            if (appConfigs.subscribeToCandles) retrieveMarketCandles()
```
prashanDYDX, line 1617: Should we just move this check inside `retrieveMarketCandles()`? (can rename to `maybeRetrieveMarketCandles()` for clarity) Would prevent someone from accidentally adding a call somewhere else without the check.

aforaleka: yee! i think we can also align with whether to adopt this pattern for the other functions for v2 as well -- but will leave that for now and keep this small

PR #272, opened 2024-03-28
1.6.25: Expose commitTriggerOrders and triggerOrdersPayload endpoints

moo-onthelawn:
TriggerOrdersInput
- Added `summary` which currently just contains the calculated `price` (limit price for limit orders, or triggerPrice + slippagePercentage for market orders)

TriggerOrdersInputCalculator
- Added calculations for calculating `summary.price` for limit and market orders
  - Currently using the same slippage percentages from `TradeInput` (ranges from 5 - 20%, confirmed with Se)
- Updated calculator to auto-deduce the `side` and `type` of an order based on its price inputs (limit order if non-null `price.limitPrice`) and position
- Updated calculator to remove `percentDiff` calculations after discussion with product - we're not sure of the best formula yet
- Updated `usdcDiff` calculation to take into account the selected orderSize

Create and expose a `triggerOrdersPayload` and `commitTriggerOrders` function
- `commitTriggerOrders(): HumanReadableTriggerOrdersPayload?` 
    - currently returns a list of placed and cancelled orders (if any), to be used by FE for local notifications (toasts)
    - I took out the logic for short term orders here since SL and TP are both stateful 
- `triggerOrdersPayload(): HumanReadableTriggerOrdersPayload?`
    - this function holds the logic to determine whether we need to create, update (cancel + create) or cancel SL/TP orders
        - it has some [extra logic](https://github.com/dydxprotocol/v4-abacus/pull/272/files#diff-f684fa19a17638d5b6d17dfd61dd715ab96db87920846ac4c92790e9cd0b8506R759) to do nothing if we attempt to update an order with the same params that currently exist 
        - limit orders that are replaced will be set to GTT with 28 days* 
    - either calls into helper `triggerOrderPayload`, or
        - this sets the payload with the established defaults:
            - All Orders
                 - reduceOnly = true
                 - postOnly = false
                 - price = calculated by calculator
                 - triggerPrice = input by user
                 - duration = 28D
           - SL/TP Market Orders
               - timeInForce = IOC
               - execution = IOC
           - SL/TP Limit Orders
               - timeInForce = GTT
               -  execution = Default (good til date)
       - `cancelOrderPayload`
           - this already exists 

## Additional Work
- Fixed some of the calculator calculations with price inputs - realized I need to take into account the position side to accurately depict percentage diffs
- Also realized it makes more sense to return negative / positive values (not absolute) from the FE perspective
- Added `fromSlTpDialog` to analytics tracking

## Testing
- Updated existing tests
- Added to v4TransactionTests (including v2)
- Tested locally on FE that we can (this is sort of janky, my local FE code needs to be cleaned up // loading states // toasts // etc): 
    - Create new orders
    - Update existing orders
    - Do nothing if updating = same order
    - Delete orders when null input
 *Note to test limit orders, I temporarily changed timeInForce + execution to GTT (because this isn't yet supported by protocol)

https://github.com/dydxprotocol/v4-abacus/assets/70078372/077a6bfe-3add-4f21-ac5a-93f8a81fbf19


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1034,6 +1034,100 @@ class V4StateManagerAdaptor(
...
+                            this.cancelOrderRecords.add(
+                                CancelOrderRecord(
+                                    subaccountNumber,
+                                    it.clientId,
+                                    submitTimeMs,
+                                ),
+                            )
+                        }
+                    }
+                    send(error, callback, it)
```
moo-onthelawn, line 1067: I feel like it's sort of an antipattern to be calling `send` multiple times in `commitTriggerOrders` (it's being called once for every order action, and then once at the end, to signal a success if no actions were taken). This makes for some weird FE ui too (if I'm waiting for onSuccess) - but unsure if there's a better solution here that doesn't involve making a new `TransactionType` which I'd really rather not do

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1750,6 +1757,121 @@ open class StateManagerAdaptor(
...
 
+    private fun triggerOrderPayload(triggerOrder: TriggerOrder, subaccountNumber: Int, marketId: String, size: Double): HumanReadablePlaceOrderPayload {
+        val clientId = Random.nextInt(0, Int.MAX_VALUE)
+        val type = triggerOrder.type?.rawValue ?: throw Exception("type is null")
+        val side = triggerOrder.side?.rawValue ?: throw Exception("side is null")
+
+        val price = triggerOrder.summary?.price ?: throw Exception("summary.price is null")
+        val triggerPrice = triggerOrder.price?.triggerPrice ?: throw Exception("triggerPrice is null")
+
+        val reduceOnly = true
```
aforaleka, line 1768: what's the plan for this one? do we need to feature gate stop limit / take profit limit because reduce only can't be true for GTT

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1750,6 +1757,121 @@ open class StateManagerAdaptor(
...
+            else -> throw Exception("invalid triggerOrderType")
+        }
+
+        val execution = when (triggerOrder.type) {
+            OrderType.stopMarket, OrderType.takeProfitMarket -> "IOC"
+            OrderType.stopLimit, OrderType.takeProfitLimit -> "Default"
+            else -> throw Exception("invalid triggerOrderType")
+        }
+
+        val duration = GoodTil.duration(TradeInputGoodUntil(28.0, "D")) ?: throw Exception("invalid duration")
```
aforaleka, line 1783: so they'll always default to 28 days? can we make a const somewhere?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1750,6 +1757,121 @@ open class StateManagerAdaptor(
...
+            // 5. No existing order -> nothing should be done
+
+            if (triggerOrder.orderId != null) {
+                val existingOrder = subaccount.orders?.firstOrNull { it.id == triggerOrder.orderId }
+                    ?: throw Exception("order is null")
+                if (triggerOrder.price?.triggerPrice != null) {
+                    if (!isTriggerOrderEqualToExistingOrder(triggerOrder, existingOrder, size)) {
+                        // (1) Existing order -> update
+                        cancelOrderPayloads.add(cancelOrderPayload(triggerOrder.orderId))
+                        placeOrderPayloads.add(triggerOrderPayload(triggerOrder, subaccountNumber, marketId, size))
```
aforaleka, line 1847: thoughts on using two vars `shouldCancelOrder` and `shouldPlaceOrder`, so these two lines don't have to be duplicated / potentially easier to read since it's just the same order?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1034,6 +1034,100 @@ class V4StateManagerAdaptor(
...
+
+            lastOrderClientId = null
+
+            val transactionCallback =
+                { response: String?, uiDelayTimeMs: Double, submitTimeMs: Double ->
+                    val error = parseTransactionResponse(response)
+                    if (error == null) {
+                        tracking(
+                            AnalyticsEvent.TradePlaceOrder.rawValue,
+                            ParsingHelper.merge(uiTrackingParams(uiDelayTimeMs), analyticsPayload)
```
aforaleka, line 1100: same here re tracking

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
aforaleka: umm maybe @johnqh can chime in here too. im not sure what the timeline is for supporting replacing orders / sending / canceling multiple orders in one tx, i think that would solve most of those concerns

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
moo-onthelawn: we'll be feature gating stop limit / take profit limit on the FE! (if something does end up going through, the order will just fail)

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
moo-onthelawn: yup - will always default - will make a const!

moo-onthelawn:
> Questions,
> 
> 1:I assume if you leave the input blank, it means to delete the existing order?

Yup! Will add unit tests

> 2: what's the product requirements on how to deal with execution mode?

We always use the default execution mode (IOC for IOC/market orders, and Default for GTD/limit orders)

> 3: how does user change timeInForce? For example, if I place a stop limit with GTT of 3 days, how does it work with the triggers?

They can't change time in force through the SL/TP dialog. If they go through the right trade panel flow, they can set a SL/TP with custom GTT time, but if they edit in the SL/TP dialog, the order will get deleted and replaced with an order of GTT 4w. All orders they create through the SL/TP dialog will also be created with 4w (with no ability to edit)

src/commonTest/kotlin/exchange.dydx.abacus/app/manager/v2/V4TransactionTests.kt
```diff
@@ -177,6 +203,85 @@ class V4TransactionTests : NetworkTests() {
...
+        assertEquals(2, transactionQueue?.size)
+        assertEquals(3, transactionCalledCount)
+        clearTransactions(2)
+
+        simulateNewBtcOrder() // 2 new limit orders created
+        assertEquals(2, transactionQueue?.size)
+        clearTransactions(2)
+
+        // Updating Existing Order
+        // testWebSocket?.simulateReceived(mock.accountsChannel.v4_subaccounts_update_1)
```
moo-onthelawn, line 260: hmm I having trouble making this set of tests work @johnqh because I think of how subaccounts are handled differently in v2 D: `testWebSocket?.simulateReceived(mock.accountsChannel.v4_subaccounts_update_1)` doesn't seem to quite be applying. Do you happen to know what the best way of getting this new position mocked in would be for the test?

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TriggerOrdersInputCalculator.kt
```diff
@@ -56,102 +71,176 @@ internal class TriggerOrdersInputCalculator(val parser: ParserProtocol) {
...
-                    modified.safeSet(
-                        "price.percentDiff",
-                        if (triggerPrice != null) Numeric.double.ONE.minus(triggerPrice.div(entryPrice)) else null,
+                        "usdcDiff",
+                        if (triggerPrice != null) {
+                            when (positionSide) {
+                                "long" -> absOrderSize.times(entryPrice.minus(triggerPrice))
+                                "short" -> absOrderSize.times(triggerPrice.minus(entryPrice))
+                                else -> null
+                            }
```
moo-onthelawn, line 108: cc @mike-dydx , here are the calculations - ended up commenting out percentDiff for now (and not setting it at all) until we come up with a better formula

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
moo-onthelawn: hm I tried this out and felt like it was actually harder to parse out what was happening - so leaning towards keeping it like this if that's alright with yoU!

src/commonTest/kotlin/exchange.dydx.abacus/tests/payloads/AccountsChannelMock.kt
```diff
@@ -2006,16 +2006,16 @@ internal class AccountsChannelMock {
            "type":"channel_data",
            "connection_id":"f7a21cc6-4a9f-42d7-8bbd-215f11614571",
            "message_id":119,
-           "id":"cosmos140tmxre67zelczxz2fe7j0hnzm2jwrfrl8a7vq/0",
+           "id":"cosmos1fq8q55896ljfjj7v3x0qd0z3sr78wmes940uhm/0",
```
moo-onthelawn, line 2009: @johnqh updated the mocks here to reflect the same cosmosAddress as what's used for `v4_channel_data_with_orders` and `v4accountsReceived` - lmk if this isn't right? The other tests in the database still all work

PR #283, opened 2024-04-02
Allow differing order sizes in trigger orders input

moo-onthelawn:
Add size to `TriggerOrder` so that we can allow the SL/TP dialog to handle differing SL/TP order sizes. 
- Note that `TriggerOrders.size` reflects the input size from the dialog, and will take precedence over the order size of `TriggerOrder.size`
- `TriggerOrder.size` is only filled out if there is an existing order (passed back from the FE)
- All calculations are done with `TriggerOrders.size` if it exists - otherwise, we defer to `TriggerOrder.size`
  - This logic is encapsulated in `TriggerOrdersInputCalculator.finalizeOrderFromPriceInputs` which writes the final size to `summary.size` (used in the payload for `commitTriggerOrders`) 
- Updated validation to validate against these sizes as well

Added tests, also tested locally on FE here:

https://github.com/dydxprotocol/v4-abacus/assets/70078372/52506e9d-43fe-457f-b7c2-a3038b7bf965


src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
```diff
@@ -175,18 +184,23 @@ internal class TriggerOrdersInputValidator(
             equityTier?.nextLevelRequiredTotalNetCollateralUSD
         val numOrders = orderCount(subaccount)
         var numOrdersToCreate = 0
+        var numOrdersToCancel = 0
```
moo-onthelawn, line 187: added logic to case for orders implicitly being cancelled by user's current form input

mike-dydx (approved):

PR #286, opened 2024-04-02
Combine interface

johnqh:


ruixhuang (approved):

prashanDYDX:
What's the context for this change?

ruixhuang:
> What's the context for this change?

Android has issue using multiple Abacus interfaces as variable type.  It can only be fixed in Abacus.

PR #288, opened 2024-04-03
Update place/cancel orders flow + add new analytics events

aforaleka:
**Update order events lifecycle**:
1. TradePlaceOrderClick (when user clicks button on FE)
2. TradePlaceOrder (when order gets submitted to network)
3. TradePlaceOrderSubmissionConfirmed (Tx confirmed - sync/commit)
4. TradePlaceOrderSubmissionFailed(Tx failed)
5. TradePlaceOrderConfirmed (indexer ws response with new order)
(same applies to cancels)

**Roundtrip ms overestimation fix**:
- bug when indexer responds with new order before order gets saved locally, so matching failed and confirm event was only recorded on subsequent matches (another subaccount update)
- fix: save the order during submission before tx is confirmed, to guarantee above matching will find the locally saved order

**Refactoring**
- separate tracking to their own functions
- `startTransaction` with onSubmitTransaction

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
@@ -592,7 +592,7 @@ internal class SubaccountSupervisor(
                     useTransactionQueue,
                 )
             } else {
-                helper.send(error, callback, payload)
+                helper.send(error, callback, transferPayload)
             }
         }
 


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -569,216 +540,226 @@ internal class SubaccountSupervisor(
...
                 )
             } else {
-                transactionQueue.enqueue(
-                    TransactionParams(
-                        TransactionType.PlaceOrder,
-                        string,
-                        transactionCallback,
-                    ),
-                )
+                helper.send(error, callback, payload)
```
aforaleka, line 583: should this be the transferPayload or just orderPayload? @jaredvu

aforaleka: umm seems like jared hasn't incorporated it to v4-web so yea i'll change to transfer for now.. it used to be the order payload tho

PR #289, opened 2024-04-04
1.6.30: Fix commitTriggerOrders payload return type

moo-onthelawn:
Fix commitTriggerOrders to always return a payload of `HumanReadableTriggerOrdersPayload`, rather than `HumanReadableTriggerOrdersPayload | HumanReadablePlaceOrderPayload | HumanReadableCancelOrderPayload`

Tested via logging the output via `Logger.d` locally


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1064,7 +1064,7 @@ class V4StateManagerAdaptor(
                             )
                         }
                     }
-                    send(error, callback, it)
```
prashanDYDX, line 1067: nit: for lambda blocks longer than one line, we should used named params instead of `it` (as an aside, we should really get a linter/static-analysis tool setup, there's a simple lint rule for this)

prashanDYDX (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
moo-onthelawn: ooh will fix!

PR #292, opened 2024-04-05
1.6.32: implement percentage diff calculations for sl/tp modal

moo-onthelawn:
based off of formula that: 

`percentageDiff = leverage * usdcDiff / notionalTotal`

I double checked it against Se's spreadsheet, but will also make him take a pass once I get a staged FE up


src/commonMain/kotlin/exchange.dydx.abacus/calculator/TriggerOrdersInputCalculator.kt
```diff
@@ -99,68 +101,112 @@ internal class TriggerOrdersInputCalculator(val parser: ParserProtocol) {
...
-                        } else {
-                            null
-                        },
-                    )
+                            },
+                        )
+                        modified.safeSet(
+                            "percentDiff",
+                            when (positionSide) {
+                                "long" -> leverage.times(size.times(entryPrice.minus(triggerPrice))).div(notionalTotal)
```
ruixhuang, line 120: Might be safer to check `val != 0.0` when doing `.div(val)`

moo-onthelawn: ooh good point - I think a valid position should never have 0 notional value, size or leverage - but added to a check above!

PR #295, opened 2024-04-09
feat(Compliance): query for geo and compliance/screen during onboarding

rosepuppy:
- Add `Compliance` to `PerpetualState`, it will be updated similar to `Restriction`
- Query https://api.dydx.exchange/v4/geo on readyToConnect. This is the same endpoint for all network.
- Query for `v4/compliance/screen` during onboarding, when sourceAddress is available, and also when account address become available.
- Keep `Restriction` to be backwards compatible, but it will should be cleaned up once compliance work is completed.


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -432,9 +436,11 @@ open class StateManagerAdaptor(
             }
             if (sourceAddress != null) {
                 screenSourceAddress()
+                sourceAddress?.let { complianceScreen(it) }
```
ruixhuang, line 439: Nit: Since there is already a sourceAddress != null check, shouldn't need to do a let.  Can simply call complianceScreen(sourceAddress)

ruixhuang (approved):

Issue #291, opened 2024-04-05
abacus.js not generated


bncdemo:
Hello,

I am running the command: gradlew assembleJsPackage
The build is successful, but the abacus.js is not generated.

bncdemo:
Still no abacus.js generated that the \v4-abacus-main\integration\Web\index.html could run

prashanDYDX:
Without more info it's hard to provide any help here. I'd recommend checking out the Kotlin JS documentation and https://npm-publish.petuska.dev/latest/ (this is where `assembleJsPackage` comes from).

You can checkout https://github.com/dydxprotocol/v4-web/blob/main/scripts/install-local-abacus.js to see how abacus is integrated into the v4-web repo.

prashanDYDX:
Yeah looks like those docs are a bit out of date, the next section is accurate though:
```
You can export the build to a package locally with the following commands:

cd ./build/packages/js

npm pack
```

In general, Kotlin JS is very poorly documented. :( I would get used to just digging around, running tasks, and seeing what is generated in the `build` folder.

bncdemo:
> Yeah looks like those docs are a bit out of date, the next section is accurate though:
> 
> ```
> You can export the build to a package locally with the following commands:
> 
> cd ./build/packages/js
> 
> npm pack
> ```
> 
> In general, Kotlin JS is very poorly documented. :( I would get used to just digging around, running tasks, and seeing what is generated in the `build` folder.

npm pack run without error, but I got nowhere with that.

The abacus.js file is missing, running the \v4-abacus-main\integration\Web\index.html fail with abacus.js file not found. 

Do you know how to create the abacus.js, so the \v4-abacus-main\integration\Web\index.html able to run?

prashanDYDX:
The sample is also out of date. If you look at the source for `index.html` you'll notice that it is pointing to the wrong location. 

abacusjs.js should be generated at `build/packages/js`.

prashanDYDX:
Updated README.

bncdemo:
> The sample is also out of date. If you look at the source for `index.html` you'll notice that it is pointing to the wrong location.
> 
> abacusjs.js should be generated at `build/packages/js`.
> 
> Looking at the `index.html` code, I'm honestly not even sure if it will run tbh.

I did go that route _before_ opening this issue ticket.

referencing the ../../build/packages/js/abacusjs.js file ends up with:
[...]

PR #300, opened 2024-04-10
Add input fields to the trigger order validation

ruixhuang:
so that front-end client knows which fields have caused the validation to fail.

Tested on the Android client.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
@@ -5,6 +5,7 @@ import exchange.dydx.abacus.protocols.ParserProtocol
 import exchange.dydx.abacus.state.app.helper.Formatter
 import exchange.dydx.abacus.state.manager.BlockAndTime
 import exchange.dydx.abacus.state.manager.V4Environment
+import exchange.dydx.abacus.state.model.TriggerOrdersInputField
 import exchange.dydx.abacus.utils.Rounder
 
 internal data class EquityTier(
@@ -22,16 +23,6 @@ enum class RelativeToPrice(val rawValue: String) {
         operator fun invoke(rawValue: String) =
             values().firstOrNull { it.rawValue == rawValue }
     }
-}
-
-enum class TriggerInputField(val rawValue: String) {
-    TP_TRIGGER_PRICE("TP_TRIGGER_PRICE"),
-    TP_GAIN_LOSS("TP_GAIN_LOSS"),
-    SL_TRIGGER_PRICE("SL_TRIGGER_PRICE"),
-    SL_GAIN_LOSS("SL_GAIN_LOSS"),
-    ORDER_SIZE("ORDER_SIZE"),
-    TP_LIMIT_PRICE("TP_LIMIT_PRICE"),
-    SL_LIMIT_PRICE("SL_LIMIT_PRICE"),
 }
 
 internal class TriggerOrdersInputValidator(
@@ -215,7 +206,7 @@ internal class TriggerOrdersInputValidator(
                     error(
                         "ERROR",
                         "USER_MAX_ORDERS",
-                        null,           // No input field since the error is not related to a specific field
+                        null, // No input field since the error is not related to a specific field
                         null,
                         "ERRORS.TRADE_BOX_TITLE.USER_MAX_ORDERS",
                         "ERRORS.TRADE_BOX.USER_MAX_ORDERS_FOR_CURRENT_EQUITY_TIER",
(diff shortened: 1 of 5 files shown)


prashanDYDX:
lets add some unit tests?

ruixhuang:
There is no UT on any existing validators. Let’s add after changing the
data type to static since the interfaces will change.

moo-onthelawn:
> There is no UT on any existing validators. 

Hmm there is an existing test file for this validator - [TriggerOrdersInputValidationTest](https://github.com/dydxprotocol/v4-abacus/blob/main/src/commonTest/kotlin/exchange.dydx.abacus/validation/TriggerOrdersInputValidationTests.kt)

src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
```diff
@@ -341,15 +351,39 @@ internal class TriggerOrdersInputValidator(
...
+            }
+            list
+        } else if (type == "TAKE_PROFIT" || type == "TAKE_PROFIT_MARKET") {
+            var list: MutableList<String> = mutableListOf()
+            if (triggerPrice != null && triggerPrice <= 0) {
+                list.add(TriggerInputField.TP_TRIGGER_PRICE.rawValue)
+                list.add(TriggerInputField.TP_GAIN_LOSS.rawValue)
+            }
+            if (limitPrice != null && limitPrice <= 0) {
+                list.add(TriggerInputField.TP_LIMIT_PRICE.rawValue)
```
moo-onthelawn, line 374: Instead of the `TriggerInputField` type, I wonder if, to be more in line with what `TradeInputDataValidator` [does](https://github.com/dydxprotocol/v4-abacus/blob/ed1bc7241131ce6fa3ebea97489c760aa1a2a06f/src/commonMain/kotlin/exchange.dydx.abacus/validator/trade/TradeInputDataValidator.kt#L333), we should be using things like `stopLossOrder.price.triggerPrice`, `stopLossOrder.price.limitPrice`, stopLossOrder.price.usdcDiff` etc instead?

src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
```diff
@@ -483,7 +528,7 @@ internal class TriggerOrdersInputValidator(
                             error(
                                 "ERROR",
                                 "ORDER_SIZE_BELOW_MIN_SIZE",
-                                null,
+                                listOf(TriggerInputField.ORDER_SIZE.rawValue),
```
moo-onthelawn, line 531: similarly this would just be `size`

ruixhuang:
Ahh.  Somehow I didn’t find it.

src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
```diff
@@ -341,15 +351,39 @@ internal class TriggerOrdersInputValidator(
...
         triggerOrder: Map<String, Any>,
     ): List<Any>? {
+        val type = parser.asString(triggerOrder["type"])
         val triggerPrice = parser.asDouble(parser.value(triggerOrder, "price.triggerPrice"))
         val limitPrice = parser.asDouble(parser.value(triggerOrder, "price.limitPrice"))
+        val fields = if (type == "STOP_LIMIT" || type == "STOP_MARKET") {
+            var list: MutableList<String> = mutableListOf()
+            if (triggerPrice != null && triggerPrice <= 0) {
+                list.add(TriggerInputField.SL_TRIGGER_PRICE.rawValue)
+                list.add(TriggerInputField.SL_GAIN_LOSS.rawValue)
```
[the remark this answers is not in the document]
ruixhuang: The Gain/Loss dollar and percent field can change the trigger price, right?

src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
ruixhuang: OK.. Will update.  But it's a just a string token, and once we migrate to static type it will just be an enum.

src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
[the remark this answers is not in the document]
ruixhuang: Nevermind.. just saw there is "price.input" that stores the input field.  We can just use that

bncdemo:
Is there nobody who can answer github issues?
https://github.com/dydxprotocol/v4-abacus/issues/291
Does this project capable to generate js that can be run in webbrowser? (example: \v4-abacus-main\integration\Web\index.html)

PR #302, opened 2024-04-11
v1.6.40: Fix percentage calculations, fix double payload return on commitTriggerOrders

moo-onthelawn:
This is the last time I'll fix percentage calculations; worked with Se for an afternoon and we finally feel confident on it. Missing part to the calculation was that:
- We were using the fixed notionalTotal rather than the scaled notionalTotal (scaled to user order size)
- We were missing (order) size in several calculations

None of the tests had to be updated because `size` of ETH in our tests is just 1, go figure (this is mocked out in the channels data, I don't feel super strongly about mocking some new data to test a non-1 size, but can if others feel strongly; reasoning: I don't think this calculation should change anymore/be reused anywhere else).

Also fixed double callback being returned in `commitTriggerOrders`: `send` is called from both `submitCancelOrder` and  `submitPlaceOrder`, so we should only be calling the final `send` if there were in fact, no cancel or place orders in our trigger orders payload. (Without this change, web ends up sending double notifications because we trigger notifications based on the payloads recieved back).

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
@@ -1195,7 +1195,10 @@ class V4StateManagerAdaptor(
             submitPlaceOrder(callback, it, analyticsPayload, true)
         }
 
-        send(null, callback, payloads)
+        if (payloads.cancelOrderPayloads.isEmpty() && payloads.placeOrderPayloads.isEmpty()) {
+            send(null, callback, payloads)
+        }
+
         return payloads
     }
 
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
@@ -783,7 +783,10 @@ internal class SubaccountSupervisor(
             submitPlaceOrder(callback, payload, analyticsPayload, true)
         }
 
-        helper.send(null, callback, payloads)
+        if (payloads.cancelOrderPayloads.isEmpty() && payloads.placeOrderPayloads.isEmpty()) {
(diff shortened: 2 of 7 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1195,7 +1195,10 @@ class V4StateManagerAdaptor(
             submitPlaceOrder(callback, it, analyticsPayload, true)
```
mike-dydx, line 1195: might we need to remove the callback here and line 1186?

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -783,7 +783,10 @@ internal class SubaccountSupervisor(
             submitPlaceOrder(callback, payload, analyticsPayload, true)
         }
 
-        helper.send(null, callback, payloads)
+        if (payloads.cancelOrderPayloads.isEmpty() && payloads.placeOrderPayloads.isEmpty()) {
+            helper.send(null, callback, payloads)
+        }
```
mike-dydx, line 788: what is the case when both of these are empty?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
moo-onthelawn: hmm yeah so that's accurate - right now there's a callback for _every_ cancel and place order message. `submitPlaceOrder` is currently a shared function between the trade form flows and trigger form flows. If we wanted to only send back a single callback for the aggregate order messages we'd have to create a new function that only calls `send` once - though I'm unsure about the details of when to send `error` vs no error (what if one transaction fails but the other passes)?

mike-dydx (approved):

PR #309, opened 2024-04-22
Fix trigger order placement params

moo-onthelawn:
was mildly confused why trigger limit orders were successfully going through the backend, only to realize that the params I was sending back were wrong.

I thought trigger orders had a `timeInForce` of IOC (market) or GTT (limit), but turns out that assumption is incorrect after playing around with the existing stop loss / take profit orders (and further reading the code).
- Time in Force is only [non null](https://github.com/dydxprotocol/v4-abacus/blob/main/src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt#L865) on limit orders (excluding trigger limit orders) which means my logic for setting TIF was wrong (should always be null)

Also realized that execution being set to `Default` throughout `StateManagerAdaptor` and `SubaccountSupervisor` was wrong, after hitting some weird errors in v4-client; it's because it [expects](https://github.com/dydxprotocol/v4-client-public-snapshot/blob/9c01e097d9146988cc253b775781ef87f2790fc5/src/clients/constants.ts#L81) `DEFAULT`. This didn't hit before in the existing place order / close position code because we never hit this edge case.

## Testing
- Verified that things are now correct by building locally in web and logging (in abacus) the time in force and execution for orders submitted through the trade panel and SL/TP modal (and confirming they matched). 
- Verified that limit orders submitted through SL/TP modal are failing with broadcast error [9003](https://github.com/dydxprotocol/v4-chain-research/blob/f35be519d569e753475a2ae36f0bf323af80abbe/protocol/x/clob/types/errors.go#L483) (ErrReduceOnlyDisabled)
Screenshot 2024-04-22 at 4 05 38 PM

- Also confirmed I didn't break any existing things with the pivot of "Default" -> "DEFAULT" by testing "close position" (and verifying it still works).
- Verified that non limit SL/TP orders still work

https://github.com/dydxprotocol/v4-abacus/assets/70078372/007704bf-2515-4a33-9331-79c52a33b340


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1802,15 +1802,11 @@ open class StateManagerAdaptor(
...
 
-        val timeInForce = when (triggerOrder.type) {
-            OrderType.stopMarket, OrderType.takeProfitMarket -> "IOC"
-            OrderType.stopLimit, OrderType.takeProfitLimit -> "GTT"
-            else -> throw Exception("invalid triggerOrderType")
-        }
+        val timeInForce = null;
 
         val execution = when (triggerOrder.type) {
             OrderType.stopMarket, OrderType.takeProfitMarket -> "IOC"
```
mike-dydx, line 1808: this could also be FOK or IOC here. I wonder if this should be set to DEFAULT as well? Unsure if that's supported for non-limit TP/SL orders

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1802,15 +1802,11 @@ open class StateManagerAdaptor(
...
-            OrderType.stopLimit, OrderType.takeProfitLimit -> "Default"
+            OrderType.stopLimit, OrderType.takeProfitLimit -> "DEFAULT"
```
prashanDYDX, line 1809: Should these instead be an enum also?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1802,15 +1802,11 @@ open class StateManagerAdaptor(
         val reduceOnly = true
         val postOnly = false
 
-        val timeInForce = when (triggerOrder.type) {
-            OrderType.stopMarket, OrderType.takeProfitMarket -> "IOC"
-            OrderType.stopLimit, OrderType.takeProfitLimit -> "GTT"
-            else -> throw Exception("invalid triggerOrderType")
-        }
+        val timeInForce = null;
```
prashanDYDX, line 1805: might be helpful for future readers to explain why this is always null?

prashanDYDX:
tests?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
moo-onthelawn: I've been curious about this in the abacus codebase - we seem to use strings everywhere (instead of enums), not sure if there's a historical reason for this (was just pattern matching)

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
moo-onthelawn: ah so product confirmed that we want all non-limit TP/SL orders to have execution "IOC" (users can't change this)

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
prashanDYDX: Be the change :) From a brief look probably fine to keep as is in this PR, but would be good to perhaps explore updating some or all of the `String` fields in `HumanReadablePlaceOrderPayload` (TIF, execution, type, side).

moo-onthelawn: will explore in a follow up!

prashanDYDX (approved):

PR #311, opened 2024-04-24
Call compliance API to update compliance state during onboarding

rosepuppy:


src/commonMain/kotlin/exchange.dydx.abacus/output/Compliance.kt
```diff
@@ -19,6 +19,21 @@ enum class ComplianceStatus(val rawValue: String?) {
...
+    ONBOARD("ONBOARD"),
+    CONNECT("CONNECT"),
+    VALID_SURVEY("VALID_SURVEY"),
+    INVALID_SURVEY("INVALID_SURVEY"),
+    ;
+
+    companion object {
+        operator fun invoke(rawValue: String?) =
+            ComplianceStatus.values().firstOrNull { it.rawValue == rawValue }
+    }
```
prashanDYDX, line 34: kotlin tip: Don't actually need to have a `rawValue` and a custom method to parse. Kotlin will automatically convert the enum type <> String:
https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants

Basically you can remove the string values from the enum's constructor, then use `ComplianceAction.valueOf(rawValue)`

prashanDYDX:
Let's write some tests :) Happy to help brainstorm ways to make this more testable if unclear.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -2398,29 +2417,93 @@ open class StateManagerAdaptor(
         }
     }
 
-    open fun complianceCheck() {
+    open fun handleComplianceResponse(response: String?, httpCode: Int, callback: ((ComplianceStatus) -> Unit)?) {
```
prashanDYDX, line 2420: why are these functions open?

prashanDYDX: It's also not clear to me why this function needs a callback? Can't we just return normally?

rosepuppy: >why are these functions open?
Don't think it needs to be open 🤔 I was following the pattern for other screen functions here.

>It's also not clear to me why this function needs a callback? Can't we just return normally?
I used callback because I want to call the geoblock endpoint after getting the response from the screen endpoint. Let me know if there is a better way to do this!

src/commonMain/kotlin/exchange.dydx.abacus/output/Compliance.kt
rosepuppy: do you mean just like:
`enum ComplianceAction { ONBOARD, CONNECT, ... }`

I am following the pattern for enums at other places I wonder if there is a reason we always use `rawValue`

detekt-baseline.xml
```diff
@@ -436,6 +436,7 @@
     <ID>InvalidPackageDeclaration:MarketsProcessor.kt$package exchange.dydx.abacus.processor.markets</ID>
     <ID>InvalidPackageDeclaration:MarketsSummaryProcessor.kt$package exchange.dydx.abacus.processor.markets</ID>
     <ID>InvalidPackageDeclaration:MarketsSupervisor.kt$package exchange.dydx.abacus.state.v2.supervisor</ID>
+    <ID>InvalidPackageDeclaration:Address.kt$package exchange.dydx.abacus.state.manager.utils</ID>
```
prashanDYDX, line 439: Why are we adding to the baseline here? We should just fix this?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -2398,28 +2416,103 @@ open class StateManagerAdaptor(
...
-                    compliance = if (success(httpCode) && response != null) {
-                        val payload = parser.decodeJsonObject(response)?.toIMap()
-                        if (payload != null) {
-                            val status = parser.asString(payload["status"])
-                            val complianceStatus = ComplianceStatus.invoke(status) ?: ComplianceStatus.UNKNOWN
-                            Compliance(compliance?.geo, complianceStatus)
-                        } else {
-                            Compliance(compliance?.geo, ComplianceStatus.UNKNOWN)
+                    handleComplianceResponse(response, httpCode)
+                    val res = parser.decodeJsonObject(response)?.toIMap()
```
prashanDYDX, line 2509: I'm confused, why are we duplicating our decoding? Why can't we return `compliance` in `handleComplianceResponse` and then use it here?

rosepuppy: oh good point, I'll do that

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -2398,29 +2416,98 @@ open class StateManagerAdaptor(
...
-                            val complianceStatus = ComplianceStatus.invoke(status) ?: ComplianceStatus.UNKNOWN
-                            Compliance(compliance?.geo, complianceStatus)
-                        } else {
-                            Compliance(compliance?.geo, ComplianceStatus.UNKNOWN)
-                        }
-                    } else {
-                        Compliance(compliance?.geo, ComplianceStatus.UNKNOWN)
-                    }
+                    val complianceStatus = handleComplianceResponse(response, httpCode)
+                    updateCompliance(address.rawAddress, complianceStatus)
```
prashanDYDX, line 2510: i think we forgot to check `DydxAddress` vs `EvmAddress`? To enforce this, we can change the signature of `updateCompliance()` to take in `DydxAddress` instead of `String`.

prashanDYDX (approved):

PR #313, opened 2024-04-24
Update stop market order liquidity warning + params

moo-onthelawn:
As per [product](https://linear.app/dydx/issue/CT-779/remove-warning-for-users-placing-stop-market-orders), update stop market slippages from:
- Major markets: 0.05 -> 0.1
- Non-major markets: 0.1 -> 0.2

For cleanliness, updated constants to be shared between `TriggerOrdersInputCalculator` and `TradeInputCalculator`; removed unused `MARKET_ORDER_SLIPPAGE_WARNING_THRESHOLD`.

Also remove `STOP_MARKET_ORDER_MAY_NOT_EXECUTE ` warning prompting users to use stop limit orders.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
@@ -1,6 +1,11 @@
 package exchange.dydx.abacus.calculator
 
 import abs
+import exchange.dydx.abacus.calculator.SlippageConstants.MARKET_ORDER_MAX_SLIPPAGE
+import exchange.dydx.abacus.calculator.SlippageConstants.STOP_MARKET_ORDER_SLIPPAGE_BUFFER
+import exchange.dydx.abacus.calculator.SlippageConstants.STOP_MARKET_ORDER_SLIPPAGE_BUFFER_MAJOR_MARKET
+import exchange.dydx.abacus.calculator.SlippageConstants.TAKE_PROFIT_MARKET_ORDER_SLIPPAGE_BUFFER
+import exchange.dydx.abacus.calculator.SlippageConstants.TAKE_PROFIT_MARKET_ORDER_SLIPPAGE_BUFFER_MAJOR_MARKET
 import exchange.dydx.abacus.protocols.ParserProtocol
 import exchange.dydx.abacus.state.manager.EnvironmentFeatureFlags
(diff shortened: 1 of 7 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
```diff
@@ -26,6 +26,14 @@ enum class TradeCalculation(val rawValue: String) {
     }
 }
 
+object SlippageConstants {
+    internal const val MARKET_ORDER_MAX_SLIPPAGE = 0.05
```
prashanDYDX, line 30: nit, can do `internal object` and drop all the `internal` from the fields inside

src/commonTest/kotlin/exchange.dydx.abacus/validation/TradeRequiredInputTests.kt
```diff
@@ -213,12 +213,7 @@ class TradeRequiredInputTests : V3BaseTests() {
                         "trade": {
                             "type": "STOP_MARKET"
                         },
-                        "errors": [
-                            {
-                                "type": "WARNING",
-                                "code": "STOP_MARKET_ORDER_MAY_NOT_EXECUTE"
-                            }
-                        ]
+                        "errors": null
```
yogurtandjam, line 216: [q] is this supposed to be an unhappy path test? if so should we change the test input params to ensure this still throws an error?

prashanDYDX (approved):

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
moo-onthelawn: so much kotlin syntax im learning ty!!

src/commonTest/kotlin/exchange.dydx.abacus/validation/TradeRequiredInputTests.kt
moo-onthelawn: eh it wasn't really an unhappy path - it was just a warning (non blocking for a user; errors are blocking) that we ripped out. I think I'm fine leaving it like this for now; we could add a test for `SELL_TRIGGER_TOO_CLOSE_TO_LIQUIDATION_PRICE` but I didn't touch that logic (and the mock data doesn't currently have a liquidation price) so inclined to leave it like this for now!

PR #315, opened 2024-04-25
v1.6.48: Add liquidation warning for stop market trigger price

moo-onthelawn:
Tested by building locally on v4-web; added tests. Logic mainly copied from [here](https://github.com/dydxprotocol/v4-abacus/blob/main/src/commonMain/kotlin/exchange.dydx.abacus/validator/trade/TradeTriggerPriceValidator.kt#L91). before pulling in this updated abacus, web/mobile should also pull in https://github.com/dydxprotocol/v4-localization/pull/411/.

One nit about the current localization:
1. There's a double $ sign because the localization adds an unnecessary `$` (and we format it as a price) - decided it was better to remove formatting from abacus and leave it to internationalization team (esp since we're not guaranteed that $ comes before the number in diff languages) 

Screenshot 2024-04-25 at 1 16 53 PM


src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
```diff
@@ -543,6 +621,19 @@ internal class TriggerOrdersInputValidator(
         return null
     }
 
+    private fun requiredTriggerToLiquidationPrice(type: String?, side: String?): RelativeToPrice? {
+        return when (type) {
+            "STOP_MARKET" ->
```
prashanDYDX, line 626: If we make these case enums instead of strings, we can avoid having an `else -> null` case and remove a source of nullability.

I would also consider whether `else` is even a valid state to be in. If not, we should just `error()` out instead of returning null. In general, we should be looking to avoid nullable types as much as possible.

src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
```diff
@@ -166,9 +169,84 @@ internal class TriggerOrdersInputValidator(
...
+                                    "tickSize" to tickSize,
+                                ),
+                            ),
+                        ),
+                    )
+                } else {
+                    null
+                }
+            }
+            else -> null
```
prashanDYDX, line 246: is this else case needed...? and if it is, i would again consider whether this is a valid state to be in and if we should just error out instead.

prashanDYDX: with `enum` and `sealed` types, the compiler can enforce that we are checking every single case, so we shouldn't need an `else` case here.

src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
```diff
@@ -130,6 +132,7 @@ internal class TriggerOrdersInputValidator(
         market: Map<String, Any>?,
         oraclePrice: Double,
         tickSize: String,
+        liquidationPriceErrors: List<Any>? = null,
```
mike-dydx, line 135: might we be able to instead pass `position` as an argument and then move the liquidationPriceErrors calculations into this function body? Seems like that is the existing pattern for this function's body

moo-onthelawn: hmm so liquidationPriceErrors only applies to stop loss orders (not take profit) - which is why I have it passed in from the caller. I could pass in a `isStopLoss?` as a param here but not sure if that's much better 🤔

src/commonMain/kotlin/exchange.dydx.abacus/validator/trade/TradeTriggerPriceValidator.kt
```diff
@@ -234,12 +236,13 @@ internal class TradeTriggerPriceValidator(
     private fun triggerToLiquidationError(
         triggerToLiquidation: RelativeToPrice,
         triggerLiquidation: Double,
+        tickSize: String,
     ): Map<String, Any> {
         val fields = listOf("price.triggerPrice")
         val action = "APP.TRADE.MODIFY_TRIGGER_PRICE"
         // Localizations uses TRIGGER_PRICE_LIMIT as paramater name
         val params =
-            mapOf("TRIGGER_PRICE_LIMIT" to mapOf("value" to triggerLiquidation, "format" to "price"))
```
moo-onthelawn, line 242: got rid of formatting to price because I don't think this is actually great for internationalization - the $ sign can go before or after the number depending on the language, and I don't think the formatter actually takes into account language? (maybe I'm wrong - but it feels like we should punt this responsibility to our internationalization team)

src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
```diff
@@ -166,9 +166,88 @@ internal class TriggerOrdersInputValidator(
...
+                } else {
+                    null
+                }
+            }
+            RelativeToPrice.BELOW -> {
+                if (triggerPrice >= liquidationPrice) {
+                    liquidationPriceError(
+                        "BUY_TRIGGER_TOO_CLOSE_TO_LIQUIDATION_PRICE",
+                        "ERRORS.TRIGGERS_FORM_TITLE.BUY_TRIGGER_TOO_CLOSE_TO_LIQUIDATION_PRICE",
+                        "ERRORS.TRIGGERS_FORM.BUY_TRIGGER_TOO_CLOSE_TO_LIQUIDATION_PRICE_NO_LIMIT",
```
moo-onthelawn, line 214: strings localized here: https://github.com/dydxprotocol/v4-localization/pull/411/

src/commonMain/kotlin/exchange.dydx.abacus/validator/TriggerOrdersInputValidator.kt
moo-onthelawn: ah interesting; yeah the current pattern I've seen in the repo is to return early (return null) if we get an unexpected null type (but then detekt yelled at me because we had too many return statements in a function oops). Happy to pivot to using `error()` more liberally - I think this whole file could probably be revamped a bit more (i basically did the min to get it to pass detekt) - filing a tix [here](https://linear.app/dydx/issue/CT-786/clean-up-triggerordersinputvalidatorkt-file).

mike-dydx (approved):

src/commonMain/kotlin/exchange.dydx.abacus/validator/trade/TradeTriggerPriceValidator.kt
moo-onthelawn: rip actually added it back because tickSize doesn't work unless the format is "price". Will go back to localization repo and remove the extra `$`

PR #319, opened 2024-04-26
Refactoring NotificationsProvider

ruixhuang:
No new code, just breaking down NotificationsProvider to multiple provider objects.   I am going to add notification logic for the TP/SL trigger order status, so it's better to refactor the code first.


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+
+class NotificationsProvider(
+    private val uiImplementations: UIImplementations,
+    private val environment: V4Environment,
+    private val parser: ParserProtocol,
+    private val jsonEncoder: JsonEncoder,
+    private val useParentSubaccount: Boolean = false,
+) : NotificationsProviderProtocol {
+
+    private val providers = listOf(
```
prashanDYDX, line 37: If we pass these into the constructor, then `NotificationsProvider` itself does not need to worry about the transitive dependencies, and will make `NotificationsProvider` easier to test.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/providers/BlockRewardNotificationProvider.kt
```diff
@@ -0,0 +1,87 @@
...
+                        it,
+                    )
+                }
+            }
+        }
+        return notifications
+    }
+
+    private fun createBlockRewardNotification(
+        stateMachine: TradingStateMachine,
```
prashanDYDX, line 51: we're not using this param?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+import exchange.dydx.abacus.state.model.TradingStateMachine
+import exchange.dydx.abacus.utils.IMap
+import exchange.dydx.abacus.utils.JsonEncoder
+import exchange.dydx.abacus.utils.ParsingHelper
+import exchange.dydx.abacus.utils.UIImplementations
+import kollections.toIMap
+
+interface NotificationsProviderProtocol {
+    fun buildNotifications(
+        stateMachine: TradingStateMachine,
```
prashanDYDX, line 18: I think using the full `TradingStateMachine` type will make it really annoying to test each notification provider in isolation.

Rather than passing it into the method, each provider should just take in the dependencies it needs in its constructor. Then, we can create lighter-weight interfaces for `TradingStateMachine` to implement that will allow us to easily make fakes to test against.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+        subaccountNumber: Int
+    ): IMap<String, Notification>
+
+    fun asset(stateMachine: TradingStateMachine, marketId: String): Asset? {
```
prashanDYDX, line 22: Why is this method here? This seems like a fairly generic util method that could be placed somewhere else?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/providers/FillsNotificationProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+            marketImageUrl,
+            title,
+            text,
+            null,
+            paramsAsJson,
+            fill.createdAtMilliseconds,
+        )
+    }
+
+    private fun market(stateMachine: TradingStateMachine, marketId: String): PerpetualMarket? {
```
prashanDYDX, line 234: Perhaps we should just have a file for `TradingStateMachine` extension functions to easily access parts of the state:
```
fun TradingStateMachine.market(marketId: String): PerpetualMarket? = state?.market(marketId)
```

prashanDYDX:
This seems like a great opportunity to add tests. If adding tests is difficult/annoying, I think that's a code smell that there are better ways to architect our objects (pointed out some potential improvements in my other comments) that we should address now since we're refactoring anyways. Happy to help brainstorm here.

integration/iOS/Pods/Pods.xcodeproj/project.pbxproj
```diff
@@ -402,6 +373,25 @@
...
+		};
+		90EDD76926101DC4F3EA3CB210AD782B /* abacus */ = {
+			isa = PBXGroup;
+			children = (
+				FE925A01FE4CC24BC13D7294AA9341F0 /* Frameworks */,
+				F91A9C8581D5144BABA03A418A7A1BBB /* Pod */,
+				DEF004F52F2F2D30519CC77F4A88FEF2 /* Support Files */,
+			);
+			name = abacus;
+			path = "/Users/ruihuang/v4-abacus";
```
prashanDYDX, line 392: hmm do we want this checked in?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/providers/OrderStatusChangesNotificationProvider.kt
```diff
@@ -0,0 +1,151 @@
...
+                    "ORDER_TYPE" to orderType,
+                    "ORDER_TYPE_TEXT" to orderTypeText,
+                    "ORDER_STATUS" to order.status.rawValue,
+                ).filterValues { it != null } as Map<String, String>
+                ).toIMap()
+            val paramsAsJson = jsonEncoder.encode(params)
+
+            val title =
+                uiImplementations.localizer?.localize("$statusNotificationStringKey.TITLE")
+                    ?: return null
```
prashanDYDX, line 122: Probably want guard clauses at the top of the method

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+import exchange.dydx.abacus.utils.JsonEncoder
+import exchange.dydx.abacus.utils.ParsingHelper
+import exchange.dydx.abacus.utils.UIImplementations
+import kollections.toIMap
+
+interface NotificationsProviderProtocol {
+    fun buildNotifications(
+        stateMachine: TradingStateMachine,
+        subaccountNumber: Int
+    ): IMap<String, Notification>
```
prashanDYDX, line 20: I think I'm missing some context: why are notifications modeled as a Map and not a List?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
ruixhuang: The dependency is TradingStateMachine :-). We can pass it in from the constructor, but then the callers will have be updated which increases the scope of this PR.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
ruixhuang: The map is used for post-processing of the messages. See consolidateNotifications()

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
ruixhuang: This only used by the child providers, so probably no need to expose it outside.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/providers/BlockRewardNotificationProvider.kt
ruixhuang: The code is copy-paste from old NotificationsProvider.  But we can delete this.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/providers/FillsNotificationProvider.kt
ruixhuang: It's only used by this provider.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
prashanDYDX: I don't think adding a method to an interface is the right way to go about this. Interfaces should only expose methods for consumers, not helper functions for implementations.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
prashanDYDX: Passing it into the constructor allows the different implementations to more narrowly specify what they need from TradingStateMachine, rather than taking in the whole object which will make it easier to test. 

I don't think the signature updates are a significant scope increase - there are very few usages.

ruixhuang: The dependency is already passed from the constructor. 

NotificationProvider is designed to be a composite of all providers which have a common interface.  Each provider just operates on the interface, and you can write tests independently.  Without a common interface, tests will be more difficult to write since the logic is now split between the individual provider and composite provider.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
ruixhuang: OK.. Updated

ruixhuang:
> This seems like a great opportunity to add tests. If adding tests is difficult/annoying, I think that's a code smell that there are better ways to architect our objects (pointed out some potential improvements in my other comments) that we should address now since we're refactoring anyways. Happy to help brainstorm here.

This PR is intended to do a minimal refactor to unblock upcoming feature work, so we are relying on existing tests.  If additional tests are needed or they should be architected differently, please create tickets so we can prioritize them in the future.

prashanDYDX:
> > This seems like a great opportunity to add tests. If adding tests is difficult/annoying, I think that's a code smell that there are better ways to architect our objects (pointed out some potential improvements in my other comments) that we should address now since we're refactoring anyways. Happy to help brainstorm here.
> 
> This PR is intended to do a minimal refactor to unblock upcoming feature work, so we are relying on existing tests. If additional tests are needed or they should be architected differently, please create tickets so we can prioritize them in the future.

Oh okay great, I didn't realize there were already tests for this logic.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+}
+
+class NotificationsProvider(
+    private val stateMachine: TradingStateMachine,
+    private val uiImplementations: UIImplementations,
+    private val environment: V4Environment,
+    private val parser: ParserProtocol,
+    private val jsonEncoder: JsonEncoder,
+    private val useParentSubaccount: Boolean = false,
+    private val providers: List<NotificationsProviderProtocol> = listOf(
```
prashanDYDX, line 28: I would remove this default value, so that we can remove the transitive dependencies from `NotificationsProvider` (`stateMachine`, `uiImplementations`, `environment`, `jsonEncoder`, `parser`, etc)

detekt.yml
```diff
@@ -20,3 +20,13 @@ complexity:
     threshold: 100 # up from 60
   TooManyFunctions:
     active: false
+  ReturnCount:
+    active: false
+  MaxLineLength:
```
prashanDYDX: ugh i want to keep this, but i think we need to change from 4 space indent to 2 first.

prashanDYDX (approved):

PR #323, opened 2024-04-29
MOB-464 Add trigger order notification logic to Abacus

ruixhuang:
An optional PresentationProtocol can be supplied by the client for Abacus to display toast notification.  Added the trigger order notification use-case for mobile (it's turned off now for web since web FE does this logic on its own).


src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
```diff
@@ -0,0 +1,380 @@
...
+import exchange.dydx.abacus.utils.IList
+import exchange.dydx.abacus.utils.Logger
+
+interface TriggerOrderToastGeneratorProtocol {
+    fun onTriggerOrderSubmitted(subaccountNumber: Int, payload: HumanReadableTriggerOrdersPayload, state: PerpetualState?)
+    fun onTriggerOrderResponse(subaccountNumber: Int, successful: Boolean, error: ParsingError?, data: Any?)
+}
+
+class TriggerOrderToastGenerator(
+    private val presentation: PresentationProtocol?,
```
prashanDYDX, line 30: why is this nullable?

src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
```diff
@@ -0,0 +1,380 @@
...
+    private var orderPayloadStatus: MutableMap<String, TriggerOrderStatus> = mutableMapOf()
+
+    private var state: PerpetualState? = null
+
+    override fun onTriggerOrderSubmitted(
+        subaccountNumber: Int,
+        payload: HumanReadableTriggerOrdersPayload,
+        state: PerpetualState?,
+    ) {
+        if (presentation == null) {
```
prashanDYDX, line 51: Perhaps instead of taking in a nullable `PresentationProtocol`, we could just make a no-op implementation of the generator?

src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
```diff
@@ -0,0 +1,380 @@
...
+        return toasts
+    }
+
+    private fun generateForCancelOrder(
+        cancelOrderPayload: HumanReadableCancelOrderPayload,
+        takeProfitOrders: IList<SubaccountOrder>?,
+        stopLossOrders: IList<SubaccountOrder>?,
+        tickSize: String?,
+    ): Toast? {
+        if (formatter == null || localizer == null) {
```
prashanDYDX, line 165: Why even support passing in a nullable `formatter` or `localizer`? These should never be null right?

src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
```diff
@@ -0,0 +1,380 @@
...
+            null
+        }
+    }
+
+    private fun showToasts(toasts: List<Toast>) {
+        if (threading == null) {
+            Logger.e { "Threading is null" }
+            return
+        }
+        threading?.async(ThreadingType.main) {
```
prashanDYDX, line 374: hmmm it's not clear to me why this is needed, can you explain the thought process? shouldn't need to force this on main thread.

ruixhuang: To force UI update on main thread.

prashanDYDX: that's unnecessary to enforce inside of abacus. that should be handled by each client.

specifically on ios/android, we transform these callbacks into streams anyways. by default these platforms will consume these streams on the main thread when called from SwiftUI/Compose.

ruixhuang: This method won't be going through streams.  The client will update the UI directly.

Having Abacus doing the enforcement will simplify the client code.

src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
ruixhuang: We could.. there is no clear advantage of one way or another.

PR #325, opened 2024-04-30
v1.6.53: add analytics for SL/TP

moo-onthelawn:
## Summary of Logging Added
1. User clicks submit button on SL/TP
    - We log a `TriggerOrderClick` event with 
      - `marketId`
      - `positionSize: Double`
      - `[takeProfit/stopLoss]OrderAction: 'CREATE' | 'CANCEL' | 'REPLACE' | null (if no action)`
      - `[takeProfit/stopLoss]Order[Cancel/Place]ClientId: Int`
2. `Trade[Cancel/Place]Order` logged with existing params and `fromSlTpDialog: true`
3. If successful:
   - `Trade[Cancel/Place]OrderConfirmed` logged with existing params and `fromSlTpDialog: true`
   - `Trade[Cancel/Place]OrderSubmissionConfirmed` logged with existing params and `fromSlTpDialog: true`
4. If unsuccessful, `Trade[Cancel/Place]OrderFailed` logged with existing params and `fromSlTpDialog: true`

  ## Testing
Built locally on web and logged output.

| 2x create | 2x cancel | 2x replace | 1x replace, 1x no action |
| ----- | ----- | ----- | ----- |
| Screenshot 2024-04-30 at 5 07 01 PM | Screenshot 2024-04-30 at 5 06 38 PM | Screenshot 2024-04-30 at 5 07 47 PM | Screenshot 2024-04-30 at 5 08 22 PM |

## Code

1. Add more thorough logging for SL/TP
5. Distinguish trade log events from SL/TP log events:
    - Action: Click "place orders" on SL/TP dialog
    New behavior: single `TriggerOrderClick ` fired
    Old behavior: `TradePlaceOrder` and `TradeCancelOrder` fired
    - Action: Trade submit (for logging times)
    New behavior: `TradePlaceOrder`/`TradeCancelOrder` with `fromSlTp: true`
   - Action: Trade successfully submitted (for logging times)
    New behavior: `TradePlaceOrderSubmissionConfirmed `/`TradeCancelOrderSubmissionConfirmed` with `fromSlTp: true`
   - Action: Trade failed to submit (for logging times)
    New behavior: `TradePlaceOrderSubmissionFailed `/`TradeCancelOrderSubmissionFailed` with `fromSlTp: true`
   - Action: parsing place/cancel order records
    New behavior: `TradeCancelOrderConfirmed `/`TradePlaceOrderConfirmed ` with `fromSlTp: true`
6. Create a new `triggerOrdersAnalyticsPayload`
    - Only used in the `TriggerOrderClick` event; other events continue to use `placeOrderAnalyticsPayload`/`cancelOrderAnalyticsPayload` 
    - Tracks `marketId`, `(total)PositionSize`, `stopLossOrderAction`/`takeProfitOrderAction` (REPLACE / CANCEL / CREATE), `stopLossOrder[Cancel/Place]ClientId`, `takeProfitOrder[Cancel/Place]ClientId`
7. Update existing payloads with more data for analytics
   - `PlaceOrderRecord`/`CancelOrderRecord`: added `fromSlTp` to pass to `Trade[Cancel/Place]OrderConfirmed` 
   - `HumanReadableTriggerOrdersPayload`:  add `marketId`, `positionSize` for `triggerOrdersAnalyticsPayload`
   - `HumanReadableCancelOrderPayload`: add `type`


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1880,10 +1878,15 @@ open class StateManagerAdaptor(
         val cancelOrderPayloads = mutableListOf<HumanReadableCancelOrderPayload>()
         val triggerOrders = stateMachine.state?.input?.triggerOrders
 
+        val marketId = triggerOrders?.marketId ?: throw Exception("marketId is null")
```
prashanDYDX, line 1881: kotlin tip: we can instead enforce `triggerOrders` is non-null by doing:
`val triggerOrders = requireNotNull(stateMachine.state?.input?.triggerOrders) { "triggerOrders input was null" }`

then we don't need to do null checks on it later.

src/commonMain/kotlin/exchange.dydx.abacus/utils/AnalyticsUtils.kt
```diff
@@ -1,26 +1,97 @@
...
+enum class TriggerOrderAction(val rawValue: String) {
+    REPLACE("REPLACE"),
+    CANCEL("CANCEL"),
+    CREATE("CREATE"),
+    ;
+
+    companion object {
+        operator fun invoke(rawValue: String?) =
+            TriggerOrderAction.values().firstOrNull { it.rawValue == rawValue }
+    }
```
prashanDYDX, line 23: We don't need this here - Kotlin has built-in string conversion for constant enums: https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1128,13 +1151,26 @@ class V4StateManagerAdaptor(
     private fun trackOrderSubmit(
         uiClickTimeMs: Double,
         analyticsPayload: IMap<String, Any>?,
+        isTriggerOrder: Boolean,
         isCancel: Boolean = false
     ): Double {
         val submitTimeMs = Clock.System.now().toEpochMilliseconds().toDouble()
         val uiDelayTimeMs = submitTimeMs - uiClickTimeMs
```
prashanDYDX, line 1158: I missed this before, why are we converting these timestamps to Double? Shouldn't we be using Long?

src/commonMain/kotlin/exchange.dydx.abacus/utils/AnalyticsUtils.kt
```diff
@@ -1,26 +1,97 @@
...
+                }
+            }
+        }
+
+        return iMapOf(
+            "marketId" to payload.marketId,
+            "positionSize" to payload.positionSize,
+            "stopLossOrderAction" to stopLossOrderAction?.rawValue,
+            "takeProfitOrderAction" to takeProfitOrderAction?.rawValue,
+        ) as IMap<String, Any>?
```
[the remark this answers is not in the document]
moo-onthelawn: ah good point - should be fine if I just update the return type to be `IMap<String, Any?>`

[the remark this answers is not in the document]
moo-onthelawn: oh actually I see why we have all these casts in the file - it's because `ParsingHelper.merge` expects and returns `Map<String, Any>?` instead of `Map<String, Any?>`. Given how widely this is used, it seems like a pretty beefy change to fix, so I might just keep it like this for now 😢

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
moo-onthelawn: hmm unsure - this is just the existing pattern; cc @aforaleka who may have more context (I think you added these values for analytics?)

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
moo-onthelawn: part of me wants to write a codemod to go through the whole repo to replace all of the `throwException`s

yang-dydx (approved):
i only looked at the screenshots and the output looks good from analytics perspective

PR #326, opened 2024-04-30
Create Input for Adjusting Isolated Position's Margin

jaredvu:
- Add `AdjustIsolatedMarginInput` `InputType` and introduced to `Input` state
- Add `commitAdjustIsolatedMargin` to initiate transfer between parent and child subaccounts
- Update Input docs
- Added test to check parent and child subaccounts' `quoteBalance`


src/commonMain/kotlin/exchange.dydx.abacus/calculator/AdjustIsolatedMarginInputCalculator.kt
```diff
@@ -0,0 +1,140 @@
...
+                        if (isParentSubaccount) Numeric.double.POSITIVE else Numeric.double.NEGATIVE
+                    val usdcSize = (amount ?: Numeric.double.ZERO) * multiplier
+
+                    return mapOf(
+                        "usdcSize" to usdcSize,
+                    )
+                }
+
+                else -> {
+                    return null
```
prashanDYDX, line 85: is this a valid state to be in? if not, i'd recommend just `error()`ing out here.

prashanDYDX: or even better, we could parse this into an enum and get exhaustive `when` checking here.

jaredvu: good call, parsed to string but invoked via enum class.

src/commonMain/kotlin/exchange.dydx.abacus/calculator/AdjustIsolatedMarginInputCalculator.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    private fun getModifiedTransferDelta(
+        isolatedMarginAdjustment: Map<String, Any>,
+        isParentSubaccount: Boolean,
+    ): Map<String, Double>? {
+        val type = parser.asString(isolatedMarginAdjustment["type"])
+        val amount = parser.asDouble(isolatedMarginAdjustment["amount"])
+        if (type != null) {
+            when (type) {
+                "ADD" -> {
```
yogurtandjam, line 67: [q] does kotlin support enums? would be nice to have switch cases not be string literals

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
```diff
@@ -0,0 +1,175 @@
...
+                    existing?.type != type ||
+                    existing.amount != amount ||
+                    existing.childSubaccountNumber != childSubaccountNumber ||
+                    existing.adjustIsolatedMarginInputOptions != adjustIsolatedMarginInputOptions ||
+                    existing.summary !== summary ||
+                    existing.errors !== errors ||
+                    existing.errorMessage != errorMessage ||
+                    existing.fee != fee
+                ) {
+                    AdjustIsolatedMarginInput(
```
yogurtandjam, line 157: [suggestion] use named args

src/commonMain/kotlin/exchange.dydx.abacus/calculator/AdjustIsolatedMarginInputCalculator.kt
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    private fun getModifiedTransferDelta(
+        isolatedMarginAdjustment: Map<String, Any>,
+        isParentSubaccount: Boolean,
+    ): Map<String, Double>? {
+        val type = parser.asString(isolatedMarginAdjustment["type"])
+        val amount = parser.asDouble(isolatedMarginAdjustment["amount"])
+        if (type != null) {
```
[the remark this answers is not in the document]
prashanDYDX: kotlin syntax for this is:
`val type = parser.asString(isolatedMarginAdjustment["type"]) ?: return`

src/commonMain/kotlin/exchange.dydx.abacus/calculator/AdjustIsolatedMarginInputCalculator.kt
prashanDYDX: hmm i'm not seeing an enum used here?

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
```diff
@@ -0,0 +1,175 @@
...
+
+@JsExport
+@Serializable
+enum class IsolatedMarginAdjustmentType(val rawValue: String) {
+    Add("ADD"),
+    Remove("REMOVE");
+
+    companion object {
+        operator fun invoke(rawValue: String) =
+            IsolatedMarginAdjustmentType.values().firstOrNull { it.rawValue == rawValue }
```
prashanDYDX, line 93: Can use built-in Kotlin conversion here:
```
enum class IsolatedMarginAdjustmentType {
  ADD,
  REMOVE,
}
```

When parsing: `IsolatedMarginAdjustmentType.valueOf( ... )`

(also don't need semi-colons in kotlin :) )

src/commonMain/kotlin/exchange.dydx.abacus/calculator/AdjustIsolatedMarginInputCalculator.kt
```diff
@@ -0,0 +1,136 @@
+package exchange.dydx.abacus.calculator
+
+import exchange.dydx.abacus.output.input.IsolatedMarginAdjustmentType
```
prashanDYDX, line 3: can use static imports for `Add` and `Remove` specifically to make the code easier to read below.
```
import exchange.dydx.abacus.output.input.IsolatedMarginAdjustmentType.Add
import exchange.dydx.abacus.output.input.IsolatedMarginAdjustmentType.Remove
```

src/commonMain/kotlin/exchange.dydx.abacus/calculator/SubaccountTransformer.kt
```diff
@@ -199,6 +199,29 @@ internal class SubaccountTransformer {
...
     }
 
+    internal fun applyIsolatedMarginAdjustmentToWallet(
+        wallet: Map<String, Any>,
+        subaccountNumber: Int?,
+        delta: Map<String, Double>?,
+        parser: ParserProtocol,
+        period: String
+    ): Map<String, Any> {
+        return if (delta != null) {
```
prashanDYDX, line 209: can we make `delta` here non-nullable? and move the nullability check out to the consumer?

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
```diff
@@ -0,0 +1,175 @@
...
+
+import exchange.dydx.abacus.protocols.ParserProtocol
+import exchange.dydx.abacus.utils.Logger
+import kollections.JsExport
+import kotlinx.serialization.Serializable
+
+@JsExport
+@Serializable
+data class AdjustIsolatedMarginInputOptions(
+    val needsSize: Boolean?,
```
prashanDYDX, line 11: what does a `null` needsSize mean?

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
```diff
@@ -0,0 +1,175 @@
...
+}
+
+@JsExport
+@Serializable
+data class AdjustIsolatedMarginInputSummary(
+    val crossFreeCollateral: Double?,
+    val crossMarginUsage: Double?,
+    val positionMargin: Double?,
+    val positionLeverage: Double?,
+    val liquidationPrice: Double?,
```
prashanDYDX, line 44: are these values all actually nullable?

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
```diff
@@ -0,0 +1,175 @@
...
+@Serializable
+data class AdjustIsolatedMarginInput(
+    val type: IsolatedMarginAdjustmentType,
+    val amount: String?,
+    val childSubaccountNumber: Int?,
+    val adjustIsolatedMarginInputOptions: AdjustIsolatedMarginInputOptions?,
+    val summary: AdjustIsolatedMarginInputSummary?,
+    val errors: String?,
+    val errorMessage: String?,
+    val fee: Double?,
```
prashanDYDX, line 107: same here, are there required fields we can make non-nullable?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1767,6 +1769,21 @@ open class StateManagerAdaptor(
         }
     }
 
+    fun adjustIsolatedMargin(
+        data: String?,
```
prashanDYDX, line 1773: what does it mean for data to be null here?

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+AdjustIsolatedMarginInput.kt
```diff
@@ -0,0 +1,146 @@
...
+@Serializable
+enum class AdjustIsolatedMarginInputField(val rawValue: String) {
+    Type("type"),
+    Amount("amount"),
+    ChildSubaccountNumber("childSubaccountNumber");
+
+    companion object {
+        operator fun invoke(rawValue: String) =
+            AdjustIsolatedMarginInputField.entries.firstOrNull { it.rawValue == rawValue }
+    }
```
prashanDYDX, line 27: Similarly to the other enum, we can just use built-in kotlin enum handling

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
jaredvu: I think we move to the built-in kotlin conversion we lose access to `rawValue` (which is heavily leveraged in the FE code)

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+AdjustIsolatedMarginInput.kt
```diff
@@ -0,0 +1,146 @@
...
+    changes?.let { update(it) }
+    return StateResponse(state, changes, if (error != null) iListOf(error) else null)
+}
+
+fun TradingStateMachine.validAdjustIsolatedMarginInput(
+    adjustIsolatedMargin: Map<String, Any>,
+    parentSubaccountNumber: Int?,
+    typeText: String?,
+): Boolean {
+    if (typeText == null) return false
```
prashanDYDX, line 127: if we're using a guard clause like this, can we just make `typeText` non-null?

src/commonMain/kotlin/exchange.dydx.abacus/calculator/AdjustIsolatedMarginInputCalculator.kt
jaredvu: qq: wdym by easier to read? The import would be easier to read?

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
prashanDYDX: oh the equivalent built-in is `.name`: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-enum/name.html

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
jaredvu: It is possible to unset fields from the frontend by passing  `data: null`. AFAIK this is also used to initialize the `current` input. i.e. Setting the singleton from Trade -> Transfers -> Trade

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+AdjustIsolatedMarginInput.kt
jaredvu: Yep! I can audit all the fn that rely on `type` and ensure they are non null since `AdjustIsolatedMarginInput` should default to `ADD` if `type` is null.

src/commonMain/kotlin/exchange.dydx.abacus/calculator/AdjustIsolatedMarginInputCalculator.kt
jaredvu: Wow that's pretty cool! I don't have a preference on either though. With the static import I think we would lose some readability unless you double check the enum, and what happens if there are multiple enums with the same `.name`

prashanDYDX: yeah up to you, i don't feel strongly. 

re multiple enums: you have to explicitly say what type you're casting the value to, so that ends up not being an issue. ex:
```
val type = "Add"
when(IsolatedMarginAdjustmentType.valueOf(type)) {
  Add -> ...
  Remove -> ...
}
```

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
jaredvu: Do you happen to know what happens if we try to invoke an enum from a string and it doesn't exist? The current `invoke` method will result in `null`.

i.e. Protocol/Indexer adds `TRAILING_STOP` orders but it isn't defined in our `OrderType` enums. Will invoking the enum on that string result in null?

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
jaredvu: They should be `nullable`. The summary will try to populate after each change to the input. If `amount` has yet to be filled out then is possible that some of these values are `null`.

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
jaredvu: Not sure if every platform will handle this the same way as well.

prashanDYDX: ohhh that's a good callout.

Trying this out in Kotlin Playground shows us that this will throw an error: 
https://pl.kotl.in/XH_R97P6q

I also compiled an example Kotlin snippet to JS:
```
// Kotlin
enum class TestEnum {
  ADD,
  REMOVE,
}
```
```
// compiled JS
var TestEnum_ADD_instance;
var TestEnum_REMOVE_instance;
function valueOf(value) {
  switch (value) {
    case 'ADD':
      return TestEnum_ADD_getInstance();
    case 'REMOVE':
      return TestEnum_REMOVE_getInstance();
    default:
      TestEnum_initEntries();
      THROW_IAE('No enum constant value.');
      break;
  }
}
```
So can see where it calls `THROW_IAE`

prashanDYDX: Yeah, mk seems like the current approach is the best we can do. I tried to explore if we could write a generic helper function to avoid writing this for all of our enums, but seems like it's not possible:
https://youtrack.jetbrains.com/issue/KT-42619/Add-a-static-valueOfOrNull-method-to-Enums-similar-to-valueOf

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+AdjustIsolatedMarginInput.kt
prashanDYDX: nvm on this (see discussion above)

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+AdjustIsolatedMarginInput.kt
jaredvu: hmm in this specific case `typeText` can be null. Calling the input `AdjustIsolatedMarginInput(null, null)` is valid.

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
jaredvu: I think the only non-nullable field that we want would be `type` which is already the case.

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+AdjustIsolatedMarginInput.kt
```diff
@@ -0,0 +1,146 @@
...
+import exchange.dydx.abacus.utils.mutable
+import exchange.dydx.abacus.utils.mutableMapOf
+import exchange.dydx.abacus.utils.safeSet
+import kollections.JsExport
+import kollections.iListOf
+import kotlinx.serialization.Serializable
+
+@JsExport
+@Serializable
+enum class AdjustIsolatedMarginInputField(val rawValue: String) {
```
prashanDYDX, line 19: This is purely used by clients right? so can use built-in conversion? i guess you'd run into the enum naming rule (maybe disable?)

prashanDYDX (approved):

PR #333, opened 2024-05-01
Implement triggerCompliance, also implement compliance in V2

rosepuppy:


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/StateManagerAdaptorV2.kt
```diff
@@ -429,6 +445,28 @@ internal class StateManagerAdaptorV2(
         return null
     }
 
+    private fun fetchGeo() {
+        val url = "https://api.dydx.exchange/v4/geo"
```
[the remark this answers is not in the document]
rosepuppy: Sounds good, I added to environment.
It would have the same cloudflare restriction as FE I think, so FE shouldn't be able to hit it. which will only be limited to OFAC countries after this compliance project is released.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/StateManagerAdaptorV2.kt
```diff
@@ -589,4 +633,37 @@ internal class StateManagerAdaptorV2(
             )
         }
     }
+
+    private fun didSetGeo(geo: String?) {
+        val state = stateMachine.state
+        stateMachine.state = PerpetualState(
```
prashanDYDX, line 639: kotlin tip: `data class` has a built-in `copy()` method:
`stateMachine.state = stateMachine.state.copy(compliance = ...)`

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1802,6 +1802,16 @@ open class StateManagerAdaptor(
         callback(false, V4TransactionErrors.error(null, "Not implemented"), null)
     }
 
+    internal open fun triggerCompliance(action: ComplianceAction, callback: TransactionCallback) {
```
prashanDYDX, line 1805: Similar to your other PR, what is the purpose of the `callback` here? We're not doing anything asynchronous here.

PR #334, opened 2024-05-01
v1.7.24: Add function to reclaim funds

jaredvu:
- Function looks to loop through existing subaccounts to find funds w/o positions/orders. It then initiates a transfer back to the parent subaccount.
- Fix Subaccount initialization to provide correct subaccountNumber


prashanDYDX:
Tests?

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -1296,6 +1299,91 @@ internal class SubaccountSupervisor(
...
+        val subaccountQuoteBalanceMap = subaccounts.mapValues { subaccount ->
+            // If the subaccount is the parentSubaccount, skip
+            if (subaccount.value.subaccountNumber == subaccountNumber) {
+                return@mapValues 0.0
+            }
+
+            val openPositions = subaccount.value.openPositions
+            val openOrders = subaccount.value.orders?.filter { order ->
+                val status = helper.parser.asString(order.status)
+                status == "OPEN"
```
prashanDYDX, line 1317: do we need to include `UNTRIGGERED` here?

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -1296,6 +1299,91 @@ internal class SubaccountSupervisor(
...
+                field = value
+            }
+        }
+
+    private fun pollReclaimUnutilizedFunds() {
+        reclaimUnutilizedFundsTimer = null
+        helper.ioImplementations.threading?.async(ThreadingType.main) {
+            this.reclaimUnutilizedFundsTimer = helper.ioImplementations.timer?.schedule(
+                (10.seconds).inWholeSeconds.toDouble(),
+                null,
```
prashanDYDX, line 1378: can we use this `repeat` duration instead of recursively calling this function?

jaredvu: Current recursive implementation waits for a successful callback reponse for each transfer and goes through the entire list of `childSubaccounts` then creates a fresh 10 sec timer before repeating. I don't think I can achieve the same effect with `repeat`.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -1296,6 +1299,91 @@ internal class SubaccountSupervisor(
...
+        set(value) {
+            if (field !== value) {
+                field?.cancel()
+                field = value
+            }
+        }
+
+    private fun pollReclaimUnutilizedFunds() {
+        reclaimUnutilizedFundsTimer = null
+        helper.ioImplementations.threading?.async(ThreadingType.main) {
```
prashanDYDX, line 1375: wait why are we running this on the main thread?

PR #335, opened 2024-05-02
feat: add new properties to market [OTE-296]

yogurtandjam:
Threads in new properties
`openInterestLowerCap`
`openInterestUpperCap`
so we can use them to calculate the effective IMF

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/output/Market.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/output/Market.kt
@@ -289,6 +289,8 @@ data class MarketPerpetual(
     val nextFundingAtMilliseconds: Double? = null,
     val openInterest: Double,
     val openInterestUSDC: Double,
+    val openInterestLowerCap: Double? = null,
+    val openInterestUpperCap: Double? = null,
     val line: IList<Double>?,
 ) {
     companion object {
@@ -303,6 +305,8 @@ data class MarketPerpetual(
                 val nextFundingRate = parser.asDouble(data["nextFundingRate"])
                 val nextFundingAtMilliseconds =
                     parser.asDatetime(data["nextFundingAt"])?.toEpochMilliseconds()?.toDouble()
+                val openInterestLowerCap = parser.asDouble(data["openInterestLowerCap"])
+                val openInterestUpperCap = parser.asDouble(data["openInterestUpperCap"])
                 val openInterest = parser.asDouble(data["openInterest"])
                 val openInterestUSDC = parser.asDouble(data["openInterestUSDC"])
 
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/MarketProcessor.kt
```diff
@@ -111,6 +112,8 @@ internal class MarketProcessor(parser: ParserProtocol, private val calculateSpar
         "double" to mapOf(
             "volume24H" to "volume24H",
             "openInterest" to "openInterest",
+            "openInterestLowerCap" to "openInterestLowerCap",
+            "openInterestUpperCap" to "openInterestUpperCap",
```
yogurtandjam, line 116: honestly have no idea what this is for. don't even know if i need this

src/commonMain/kotlin/exchange.dydx.abacus/output/Market.kt
```diff
@@ -324,6 +328,8 @@ data class MarketPerpetual(
                             nextFundingAtMilliseconds,
                             openInterest,
                             openInterestUSDC ?: 0.0,
+                            openInterestLowerCap,
```
prashanDYDX, line 331: nit: can we update this to use named parameters?

prashanDYDX (approved):

src/commonMain/kotlin/exchange.dydx.abacus/output/Market.kt
yogurtandjam: honestly didn't know kotlin supported that. i was actually thinking to myself "damn i wish we could use kwargs/named args"

src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/MarketProcessor.kt
yogurtandjam: ah yeah i see now, in `private fun perpetual` it's used as an input to the `transform` method. makes sense

src/commonMain/kotlin/exchange.dydx.abacus/output/Market.kt
```diff
@@ -289,6 +289,8 @@ data class MarketPerpetual(
     val nextFundingAtMilliseconds: Double? = null,
     val openInterest: Double,
     val openInterestUSDC: Double,
+    val openInterestLowerCap: Double? = null,
+    val openInterestUpperCap: Double? = null,
```
prashanDYDX, line 293: Just wanna double check - are these values actually nullable in the API response?

yogurtandjam: unfortunately, [yes](https://docs.dydx.exchange/developers/indexer/indexer_websocket), AFAIK

PR #336, opened 2024-05-02
feat: calculate effective imf

yogurtandjam:
calculates effective IMF [according to spec](https://www.notion.so/dydx/OI-Margin-Fraction-Frontend-246bb0c83c2c44df825e956215130040?pvs=4#3cb71d362ac64106a07abef9c7ca94cd).

doesn't exactly follow [suggested spec for implementation](https://www.notion.so/dydx/OI-Margin-Fraction-Frontend-246bb0c83c2c44df825e956215130040?pvs=4#d6cafb2d33574ef9a5549dcc64dc3693) because it was easier for me to reason about it this way.


src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/MarketProcessor.kt
```diff
@@ -143,9 +146,28 @@ internal class MarketProcessor(parser: ParserProtocol, private val calculateSpar
...
 
+    internal fun calculateEffectiveIMF(output: Map<String, Any>, oraclePrice: Double?): Double {
+        val baseIMF = parser.asDouble(parser.value(output, "configs.initialMarginFraction"))
+        val openInterestUSDC = parser.asDouble(parser.value(output, "perpetual.openInterestUSDC"))
+        val openInterestLowerCap = parser.asDouble(parser.value(output, "perpetual.openInterestLowerCap"))
+        val openInterestUpperCap = parser.asDouble(parser.value(output, "perpetual.openInterestUpperCap"))
+
+        // need nully checks because all properties are optional in the websocket message
+        // clean up after https://linear.app/dydx/issue/OTE-301/audit-websocket-message-types-in-indexer is done
+        if (baseIMF === null) return 1.0
```
yogurtandjam, line 161: actually not sure what i should be putting here in case baseIMF does not exist.

src/commonMain/kotlin/exchange.dydx.abacus/output/Market.kt
```diff
@@ -195,30 +197,32 @@ data class MarketConfigs(
...
-                    maintenanceMarginFraction,
-                    incrementalInitialMarginFraction,
-                    incrementalPositionSize,
-                    maxPositionSize,
-                    basePositionNotional,
-                    baselinePositionSize,
-                    candleOptions,
-                    perpetualMarketType,
-                    v4,
+                    clobPairId = clobPairId,
```
ruixhuang, line 204: Not sure if you added those by hand.. But I just found that Android Studio can do this for you with two-clicks.

src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/MarketProcessor.kt
```diff
@@ -143,9 +145,30 @@ internal class MarketProcessor(parser: ParserProtocol, private val calculateSpar
...
+        val openInterest = parser.asDouble(parser.value(output, "perpetual.openInterest"))
+        val openInterestLowerCap = parser.asDouble(parser.value(output, "perpetual.openInterestLowerCap"))
+        val openInterestUpperCap = parser.asDouble(parser.value(output, "perpetual.openInterestUpperCap"))
+
+        // need nully checks because all properties are optional in the websocket message
+        // clean up after https://linear.app/dydx/issue/OTE-301/audit-websocket-message-types-in-indexer is done
+        if (baseIMF === null) return 1.0
+        if (oraclePrice == null || openInterest == null || openInterestLowerCap == null || openInterestUpperCap == null) return baseIMF
+        val openNotional = openInterest * oraclePrice
+        val scalingFactor = (openNotional - openInterestLowerCap) / (openInterestUpperCap - openInterestLowerCap)
```
ruixhuang, line 165: Better to check for dividing by zero.

src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/MarketProcessor.kt
rosepuppy: Maybe effectiveIMF should also be null, if baseIMF is null 🤔

build.gradle.kts
```diff
@@ -49,7 +49,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.6.55"
+version = "1.7.0"
```
rosepuppy, line 52: nit: not sure if this needs to be a minor bump since we are only adding a new field, shouldn't be a breaking change I think

rosepuppy (approved):

build.gradle.kts
yogurtandjam: minor bumps are non-breaking changes~ major bumps are breaking changes 👍 

[patches are supposed to be for bugfixes](https://semver.org/)
```
MAJOR version when you make incompatible API changes
MINOR version when you add functionality in a backward compatible manner
PATCH version when you make backward compatible bug fixes
```

i notice we have a habit of using patch for backwards compatible features (which i'm fine with conforming to) but it's technically incorrect 😛

src/commonMain/kotlin/exchange.dydx.abacus/output/Market.kt
yogurtandjam: yet another reason to download android studio!

src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/MarketProcessor.kt
yogurtandjam: hmm... is that going to cause any downstream issues? i was worried if a null `maxLeverage` value would cause abacus or our UI to freak out 
FWIW this should never happen, it just needs to be here because we've typed our socket messages to have all optional attributes

yogurtandjam: nvm just saw that maxleverage can be null. updating accordingly!

PR #343, opened 2024-05-08
add validator url / network tracking in analytics

aforaleka:
image

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
@@ -1564,20 +1564,13 @@ class V4StateManagerAdaptor(
     }
 
     private fun apiStateParams(): IMap<String, Any>? {
-        val indexerTime = lastIndexerCallTime?.toEpochMilliseconds()?.toDouble()
-        val validatorTime = lastValidatorCallTime?.toEpochMilliseconds()?.toDouble()
-        val interval = if (indexerTime != null) {
-            (
-                Clock.System.now().toEpochMilliseconds()
-                    .toDouble() - indexerTime
-                )
-        } else {
-            null
-        }
+        val indexerTime = lastIndexerCallTime?.toEpochMilliseconds()
+        val validatorTime = lastValidatorCallTime?.toEpochMilliseconds()
+        val interval = indexerTime?.let { Clock.System.now().toEpochMilliseconds() - it }
         return iMapOf(
-            "lastSuccessfulIndexerRPC" to indexerTime,
-            "lastSuccessfulFullNodeRPC" to validatorTime,
-            "elapsedTime" to interval,
+            "lastSuccessfulIndexerRPC" to indexerTime?.toDouble(),
+            "lastSuccessfulFullNodeRPC" to validatorTime?.toDouble(),
+            "elapsedTime" to interval?.toDouble(),
             "blockHeight" to indexerState.blockAndTime?.block,
             "nodeHeight" to validatorState.blockAndTime?.block,
             "validatorUrl" to this.validatorUrl,
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
@@ -39,6 +39,7 @@ import exchange.dydx.abacus.utils.Logger
 import exchange.dydx.abacus.utils.Numeric
 import exchange.dydx.abacus.utils.ParsingHelper
 import exchange.dydx.abacus.utils.UIImplementations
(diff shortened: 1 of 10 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1485,39 +1485,50 @@ class V4StateManagerAdaptor(
...
 
-    private fun uiTrackingParams(interval: Double): IMap<String, Any> {
-        return iMapOf(
-            "clickToSubmitOrderDelayMs" to interval,
-        )
-    }
-
-    private fun errorTrackingParams(error: ParsingError): IMap<String, Any> {
-        return if (error.stringKey != null) {
+    private fun validatorTrackingParams(): IMap<String, Any> {
```
prashanDYDX, line 1488: nit / kotlin-tip:
```
private fun validatorTrackingParams() = validatorUrl?.let { iMapOf("validatorUrl" to it) } ?: iMapOf()
```

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/NetworkSupervisor.kt
```diff
@@ -58,11 +66,15 @@ internal open class NetworkSupervisor(
     internal open fun didSetValidatorConnected(validatorConnected: Boolean) {
     }
 
+    internal open fun didSetValidatorUrl(validatorUrl: String?) {
```
prashanDYDX, line 69: We can make `NetworkSupervisor` `abstract` instead of `open`. Then we can also make this method `abstract`.

`open` declares a concrete type that we can instantiate and want to extend. `abstract` declares an interface type. In this case, we want `abstract` because this is a function we want implementations to implement themselves.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
aforaleka: wowwwww

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/ConnectionStats.kt
```diff
@@ -245,32 +246,36 @@ internal class ConnectionStats(
         updateApiState()
     }
 
+    private fun apiStateParams(): IMap<String, Any>? {
+        val indexerTime = lastIndexerCallTime?.toEpochMilliseconds()?.toDouble()
+        val validatorTime = lastValidatorCallTime?.toEpochMilliseconds()?.toDouble()
+        val interval = if (indexerTime != null) {
```
moo-onthelawn, line 252: nit: we could probably convert this to a `let` I think

moo-onthelawn (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/ConnectionStats.kt
```diff
@@ -245,32 +246,36 @@ internal class ConnectionStats(
...
+            (
+                Clock.System.now().toEpochMilliseconds()
+                    .toDouble() - indexerTime
```
prashanDYDX, line 255: Why are we using Doubles for epoch milliseconds? Shouldn't we just keep them as Longs?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1570,27 +1563,39 @@ class V4StateManagerAdaptor(
...
-            tracking(AnalyticsEvent.NetworkStatus.rawValue, params.toIMap())
+            tracking(AnalyticsEvent.NetworkStatus.rawValue)
+        }
+    }
+
+    override fun tracking(eventName: String, params: IMap<String, Any?>?) {
+        val requiredParams = apiStateParams()
+        val mergedParams = params?.let { ParsingHelper.merge(params.filterNotNull(), requiredParams) } ?: requiredParams
+        val paramsAsString = this.jsonEncoder.encode(mergedParams)
+        this.ioImplementations.threading?.async(ThreadingType.main) {
```
prashanDYDX, line 1597: why are we running this on main thread?

aforaleka: im just copying T______T

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/ConnectionStats.kt
aforaleka: errr not sure if there's reason behind it -- can switch to Long here during the subtraction but maybe need to keep them as Doubles when sending to amplitude for consistency?

PR #344, opened 2024-05-08
chore: remove validation codepath

yogurtandjam:
remove validation codepath in favor of using BE validation.

discussion point: do we need this validator at all? do we have a comprehensive list of validation rules that the BE already covers so we can remove the corresponding FE validators?

cc: @johnqh @prashanDYDX @BrendanChou @jaredvu


yogurtandjam:
no longer removing validation here. not all validation codepaths are supported on the protocol level.

PR #346, opened 2024-05-09
v1.7.6: sort squid asset + chain list alphabetically

moo-onthelawn:
Sorted in abacus so both mobile + web get this change. 

Seemed like `SquidProcessor` was the most logical place to do this - but also open to better ideas 🤔 

## Testing
Pulled and built locally into web and verified that asset options are alphabetical
Screenshot 2024-05-09 at 12 31 36 PM


src/commonMain/kotlin/exchange.dydx.abacus/output/input/TransferInput.kt
```diff
@@ -82,6 +82,7 @@ data class DepositInputOptions(
                 return if (existing?.needsSize != needsSize ||
                     existing?.needsAddress != needsAddress ||
                     existing?.needsFastSpeed != needsFastSpeed ||
+                    existing?.exchanges != exchanges ||
```
moo-onthelawn, line 85: seemed like we just forgot to add this at some point

PR #349, opened 2024-05-10
[analytics] track order status changes

aforaleka:
- better tracking of round trip time (e.g. a short term order being filled vs. just indexed)
- also useful for other future tracking (how often is an order filled during the same FE session)

tested locally:
image

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/utils/Payloads.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/utils/Payloads.kt
@@ -1,5 +1,6 @@
 package exchange.dydx.abacus.state.manager
 
+import exchange.dydx.abacus.output.input.OrderStatus
 import exchange.dydx.abacus.utils.IList
 import kollections.JsExport
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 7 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/utils/Payloads.kt
```diff
@@ -20,6 +21,7 @@ data class PlaceOrderRecord(
     val clientId: Int,
     val timestampInMilliseconds: Double,
     val fromSlTpDialog: Boolean,
+    var lastOrderStatus: OrderStatus?,
```
aforaleka, line 24: is this.. legal? :D

moo-onthelawn: nit I wonder if we can just do `val lastOrderStatus: OrderStatus? = null`

moo-onthelawn (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/utils/Payloads.kt
aforaleka: the main issue is i can't reassign lastOrderStatus if it's `val` instead of `var`, but i can add the =null there

PR #350, opened 2024-05-13
Fix compliance action param

rosepuppy:
Compliance action param is based on wether or not user has a subaccount.

The original way of determining that was incorrect. It should be checked via the response of /v4/addresses. Use `CONNECT` if /v4/address returns a valid responsem, use `ONBOARD` otherwise.

Also support updated at as a `compliance/geoblock` response.

Updated the logic for both v1 and v2.

Tested E2E via a local build on staging and dev


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1373,10 +1367,21 @@ open class StateManagerAdaptor(
         val url = accountUrl()
         if (url != null) {
             get(url, null, null, callback = { _, response, httpCode, _ ->
-                if (success(httpCode) && response != null) {
-                    update(stateMachine.account(response), oldState)
+                val isValidResponse = success(httpCode) && response != null
+                if (isValidResponse) {
+                    response?.let { update(stateMachine.account(response), oldState) }
```
ruixhuang, line 1372: `stateMachine.account(it)` ??

rosepuppy: oh good call

ruixhuang (approved):

PR #352, opened 2024-05-13
chore: Update SubaccountPosition fields to be non-nullable.

prashanDYDX:


src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -228,31 +228,31 @@ data class SubaccountPosition(
     val id: String,
     val assetId: String,
     val side: TradeStatesWithPositionSides,
-    val entryPrice: TradeStatesWithDoubleValues?,
+    val entryPrice: TradeStatesWithDoubleValues,
     val exitPrice: Double?,
     val createdAtMilliseconds: Double?,
```
prashanDYDX, line 233: It feels like this field should always be populated, but not sure if there is a case where it wouldn't be.

ruixhuang (approved):

PR #353, opened 2024-05-14
Fix ios podspec

ruixhuang:
Latest Kotlin MM tooling added a step that causes issue with our iOS build workflow.  Updating the bump_version script to remove the step.

Also removing duplicated declaration of HistoricalPnlPeriod (it's also in Settings.kt)

### DIFF
--- a/bump_version.sh
+++ b/bump_version.sh
@@ -13,6 +13,17 @@ if [ -n "$NEW_VERSION" ]; then
   echo "Version bumped to $NEW_VERSION"
 fi
 
+#
+# Update the version in the podspec file
+#
 ./gradlew podspec
+
+# Remove the warning step about ./gradlew :generateDummyFramework, since on iOS we use CocoaPods to 
+# download the source code and build locally.
+PODSPEC_FILE="v4_abacus.podspec"
+
+# replace the line "if !Dir.exist?('build/cocoapods/framework/Abacus.framework') || Dir.empty?('build/cocoapods/framework/Abacus.framework')" with "if false"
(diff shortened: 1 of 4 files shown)


bump_version.sh
```diff
@@ -13,6 +13,17 @@ if [ -n "$NEW_VERSION" ]; then
...
+# download the source code and build locally.
+PODSPEC_FILE="v4_abacus.podspec"
+
+# replace the line "if !Dir.exist?('build/cocoapods/framework/Abacus.framework') || Dir.empty?('build/cocoapods/framework/Abacus.framework')" with "if false"
+sed -i '' "s/if !Dir.exist?('build\/cocoapods\/framework\/Abacus.framework') || Dir.empty?('build\/cocoapods\/framework\/Abacus.framework')/if false/" $PODSPEC_FILE
+
 cd integration/iOS
-pod install
\ No newline at end of file
+pod install
```
prashanDYDX, line 29: Do we need this iOS integration folder? Is it useful / up-to-date?

ruixhuang: Yes, we used that every once a while to test the integration.

src/commonMain/kotlin/exchange.dydx.abacus/protocols/PublicProtocols.kt
```diff
@@ -118,20 +118,6 @@ interface WebSocketProtocol {
     fun send(message: String)
 }
 
-@JsExport
-@Serializable
-enum class HistoricalPnlPeriod(val rawValue: String) {
```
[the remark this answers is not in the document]
ruixhuang: Yes, it's a duplicated declaration.

prashanDYDX (approved):

PR #356, opened 2024-05-14
Setup kotlin-inject

prashanDYDX:
https://github.com/evant/kotlin-inject

Sets up a simple component and provides a function to create the component and access the V2 state manager from it.


ruixhuang (approved):
Please bump the version.

PR #364, opened 2024-05-16
Static Typing: Transfer resources POC

ruixhuang:
As a proof of concept, creating an internal static typed object "InternalState" and using it to store the Transfer related resources.

Since those resources don't need further processing after fetching from Squid, we just generate the output types directly.  Other use-cases might require creating new internal types to replace the data map.

There are also other transfer data that should be converted to static typing, but this PR shows that there can be an intermediate state where the data map co-existing with static typed data.

Tested with the Android app.


src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
```diff
@@ -81,6 +82,8 @@ open class TradingStateMachine(
     private val maxSubaccountNumber: Int,
     private val useParentSubaccount: Boolean,
 ) {
+    private val internalState: InternalState = InternalState()
```
prashanDYDX, line 85: should we just take this opportunity to move away from a single monolithic state object?

ruixhuang: It doesn't hurt to have an object that aggregates different sub states.  It'd make the migration a bit easier since it's a one-to-one mapping to the state dictionary.

prashanDYDX: yeah i guess a couple thoughts: 

if we want to break down the state machine into smaller pieces, it'd be easier if the different components were exposed directly rather than through a aggregate object. but maybe we can cross that bridge when we get there.

also, it seems a bit strange that `internalState` itself is a `val`, but holds mutable sub properties. IMO, it'd be more clear if either:
a) `internalState` itself was a `var` and its sub fields were immutable
b) we just moved the sub fields to be come top-level mutable fields (basically get rid of `InternalState`)

src/commonMain/kotlin/exchange.dydx.abacus/state/internalstate/InternalState.kt
```diff
@@ -0,0 +1,5 @@
+package exchange.dydx.abacus.state.internalstate
+
+internal class InternalState(
```
prashanDYDX, line 3: this should be a data class

src/commonMain/kotlin/exchange.dydx.abacus/state/internalstate/InternalTransferInputState.kt
```diff
@@ -0,0 +1,12 @@
+package exchange.dydx.abacus.state.internalstate
+
+import exchange.dydx.abacus.output.input.SelectionOption
+import exchange.dydx.abacus.output.input.TransferInputChainResource
+import exchange.dydx.abacus.output.input.TransferInputTokenResource
+
+internal class InternalTransferInputState(
```
prashanDYDX, line 7: data class

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
ruixhuang: Updated...

I'd rather delay immutable implementation until we switch all internal states to static types.  copy() has a performance overhead, and we had run into performance issue before.  We can only benchmark it after all states have been converted.   Also, the current mutable implementation mirrors 1-to-1 to the state dictionary, so it's a relatively safer refactor.  Even so, there is already substantial complexity of finding all the usage of the fields within the code-base.  Switching to immutable at this stage would make the refactor riskier.

src/commonMain/kotlin/exchange.dydx.abacus/state/internalstate/InternalTransferInputState.kt
```diff
@@ -0,0 +1,12 @@
+package exchange.dydx.abacus.state.internalstate
+
+import exchange.dydx.abacus.output.input.SelectionOption
+import exchange.dydx.abacus.output.input.TransferInputChainResource
+import exchange.dydx.abacus.output.input.TransferInputTokenResource
+
+internal data class InternalTransferInputState(
```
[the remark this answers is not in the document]
ruixhuang: It's the state that drives the transfer input screens (i.e. deposit, withdrawal, transfer).  The name is designed to be consistent with TradeInput.

prashanDYDX (approved):

PR #367, opened 2024-05-20
indexer subaccount order type update

aforaleka:
[indexer](https://github.com/dydxprotocol/v4-chain/pull/1540/files#diff-f05888a27966099d3e948815a460446251668d224e1a48df9a99d1589bdf03d4R165) will soon send ws msg when cancel is received for non existent order, thus we'll need to update the subaccount order type to handle that scenario

those fields should only really be null when the order is already canceled, so most existing code paths can still expect those fields to be nonnull


src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -636,17 +636,17 @@ data class SubaccountOrder(
     val subaccountNumber: Int?,
     val id: String,
     val clientId: Int?,
-    val type: OrderType,
```
moo-onthelawn, line 639: can we also update the `Account.md` file?

src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
```diff
@@ -119,13 +119,13 @@ class TriggerOrderToastGenerator(
...
             selectedSubaccountPosition?.side?.current?.let { currentSide ->
-                (order.type == OrderType.takeProfitMarket || order.type == OrderType.takeProfitLimit) &&
+                (order.type == OrderType.takeProfitMarket || order.type == OrderType.takeProfitLimit) && order.side != null &&
                     order.side.isOppositeOf(currentSide)
             } ?: false
         }
         val stopLossOrders = triggerOrders?.filter { order ->
             selectedSubaccountPosition?.side?.current?.let { currentSide ->
-                (order.type == OrderType.stopMarket || order.type == OrderType.stopLimit) &&
+                (order.type == OrderType.stopMarket || order.type == OrderType.stopLimit) && order.side != null &&
```
moo-onthelawn, line 128: could we simplify this to `order.side?.isOppositeOf(currentSide)`?

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor.kt
```diff
@@ -1216,12 +1216,13 @@ class V4StateManagerAdaptor(
...
         val subaccount = stateMachine.state?.subaccount(subaccountNumber)
+        // orders should have marketId unless it has already been canceled
         val existingOrder =
-            subaccount?.orders?.firstOrNull { it.id == orderId } ?: throw ParsingException(
+            subaccount?.orders?.firstOrNull { it.id == orderId && it.marketId != null } ?: throw ParsingException(
                 ParsingErrorType.MissingRequiredData,
                 "no existing order to be cancelled for $orderId",
             )
-        val marketId = existingOrder.marketId
+        val marketId = existingOrder.marketId!!
```
moo-onthelawn, line 1225: sad that kotlin can't deduce this :'(

src/commonMain/kotlin/exchange.dydx.abacus/utils/AnalyticsUtils.kt
```diff
@@ -215,8 +215,8 @@ class AnalyticsUtils {
         return iMapOf(
             "orderId" to order.id,
             "clientId" to order.clientId,
-            "type" to order.type.rawValue,
-            "side" to order.side.rawValue,
+            "type" to order.type?.rawValue,
+            "side" to order.side?.rawValue,
             "status" to order.status.rawValue,
             "timeInForce" to order.timeInForce?.rawValue,
             "marketId" to order.marketId,
```
moo-onthelawn, line 222: hmm why don't we have to update `marketId`, `size`, `price` here as well?

moo-onthelawn (commented):
I defer to @prashanDYDX for typing things in kotlin, but overall looks gouda

src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
aforaleka: um kinda! can change to `order.side?.isOppositeOf(currentSide) == true`

src/commonMain/kotlin/exchange.dydx.abacus/utils/AnalyticsUtils.kt
aforaleka: since they'll just be undefined and we don't need to do a `?.` here!

PR #369, opened 2024-05-20
MOB-514 : add `expiresAt` at to abacus

mike-dydx:
linear: [MOB-514 : add expires at to abacus](https://linear.app/dydx/issue/MOB-514/add-expires-at-to-abacus)

- adds the `expiresAt` value to the `Compliance` object which is 7 days advanced from `updatedAt`
- adds tests for v1 and v2 abacus to test calculation of `expiresAt`
- cleans up some logic for setting compliance object

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
@@ -497,10 +497,6 @@ open class TradingStateMachine(
                 }
             }
 
-            "/v4/compliance/geoblock" -> {
-                print("compliance")
-            }
-
             else -> {
                 if (url.path.contains("/v3/historical-funding/") || url.path.contains("/v4/historicalFunding/")) {
                     changes = historicalFundings(payload)
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
@@ -441,6 +441,9 @@ open class TradingStateMachine(
         return Pair(market, resolution)
     }
 
+    /**
+     * function specifically for testing spoofed rest response processing
+     */
     fun rest(
         url: AbUrl,
         payload: String,
@@ -492,6 +495,11 @@ open class TradingStateMachine(
                 if (deploymentUri != null) {
                     changes = configurations(payload, subaccountNumber, deploymentUri)
                 }
+            }
+
+            "/v4/compliance/geoblock" -> {
+                print("compliance")
+//                this.state.compliance = compliance(payload)
             }
 
             else -> {
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
@@ -499,7 +499,6 @@ open class TradingStateMachine(
 
             "/v4/compliance/geoblock" -> {
                 print("compliance")
-//                this.state.compliance = compliance(payload)
             }
 
             else -> {
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
@@ -912,7 +912,7 @@ internal open class AccountSupervisor(
             state?.transferStatuses,
             state?.restriction,
             state?.launchIncentive,
-            compliance,
+            Compliance(state?.compliance?.geo, compliance.status, compliance.updatedAt, compliance.expiresAt),
         )
         helper.ioImplementations.threading?.async(ThreadingType.main) {
             helper.stateNotification?.stateChanged(
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
@@ -604,7 +604,7 @@ internal open class AccountSupervisor(
             if (updatedAt != null) {
                 expiresAt = try {
                     Instant.parse(updatedAt).plus(7.days).toString()
-                } catch (e: Exception) {
+                } catch (e: IllegalArgumentException) {
                     Logger.e { "Error parsing compliance updatedAt: $updatedAt" }
                     null
                 }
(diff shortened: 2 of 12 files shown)


src/commonTest/kotlin/exchange.dydx.abacus/app/manager/V4RestrictionsTests.kt
```diff
@@ -140,4 +142,77 @@ class V4RestrictionsTests {
...
+        testRest?.setResponse(
+            "https://indexer.v4staging.dydx.exchange/v4/compliance/geoblock",
+            """
+                {
+                    "status": "CLOSE_ONLY",
+                    "reason": null,
+                    "updatedAt": "2024-05-14T20:40:00.415Z"
+                }
+            """.trimIndent(),
+        )
```
mike-dydx, line 167: @johnqh does it look right that I have to set this spoofed response  twice (see below)

I have to set it twice because the test factory removes a spoofed response after the first accessing:

```
override fun get(
        url: String,
        headers: IMap<String, String>?,
        callback: RestCallback,
    ) {
        if (url.contains("env.json")) {
            callback(mock.environments.environments, 200, null)
            return
        }
        if (!url.contains("localization")) {
            requests.add(url)
        }
        val response = responses[url]
        if (response != null) {
            responses.remove(url) // <---- here
            val code = parser.asInt(response)
            if (code != null) {
                callback(null, code, null)
            } else {
                callback(response, 200, null)
            }
        } else {
            callback(null, 404, null)
        }
    }
```

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -593,25 +594,24 @@ internal open class AccountSupervisor(
...
+                    Logger.e { "Error parsing compliance updatedAt: $updatedAt" }
+                    null
+                }
             }
-        } else {
-            Compliance(compliance.geo, ComplianceStatus.UNKNOWN, compliance.updatedAt)
         }
-        return compliance.status
+        compliance = Compliance(compliance.geo, complianceStatus, updatedAt, expiresAt)
+        return complianceStatus
```
mike-dydx, line 614: @rosepuppy  this should be same functionality, just cleaned up a little so that we have less verboseness in creating the Compliance object. Same thing in the abacus v1 implementation

please ensure u see no different conditional paths

src/commonTest/kotlin/exchange.dydx.abacus/app/manager/v2/V4RestrictionsTests.kt
```diff
@@ -94,6 +96,88 @@ class V4RestrictionsTests : NetworkTests() {
...
+                    "status": "CLOSE_ONLY",
+                    "reason": null,
+                    "updatedAt": "2024-05-14T20:40:00.415Z"
+                }
+            """.trimIndent(),
+        )
+
+        stateManager.triggerCompliance(ComplianceAction.CONNECT) { successful, error, data ->
+            print("")
+        }
```
mike-dydx, line 159: @rosepuppy / @johnqh can you please see [this comment](https://github.com/dydxprotocol/v4-abacus/pull/333/files#r1587797289) in a previous PR. Is there a reason that `triggerCompliance` is with a callback? It doesn't seem to have async code:

```
    internal open fun triggerCompliance(action: ComplianceAction, callback: TransactionCallback) {
        accountAddress?.let {
            if (compliance.status != ComplianceStatus.UNKNOWN) {
                updateCompliance(DydxAddress(it), compliance.status, action)
                callback(true, null, null) //this is not an asynchronous usage of a callback
            }
        }
        callback(false, V4TransactionErrors.error(null, "No account address"), null)
    }
```

rosepuppy: hmmm good point, this callback doesn't make sense here. I guess the call back should actually be passed into from updateCompliance potentially 🤔

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
rosepuppy: Awesome, thanks for cleaning this up!!

src/commonTest/kotlin/exchange.dydx.abacus/app/manager/v2/V4RestrictionsTests.kt
mike-dydx: i will leave out of this PR for now since it's out of the OG scope

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
```diff
@@ -494,6 +497,10 @@ open class TradingStateMachine(
                 }
             }
 
+            "/v4/compliance/geoblock" -> {
+                print("compliance")
```
prashanDYDX, line 501: i think you left a debug print in here

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -912,7 +912,7 @@ internal open class AccountSupervisor(
             state?.transferStatuses,
             state?.restriction,
             state?.launchIncentive,
-            Compliance(state?.compliance?.geo, compliance.status, compliance.updatedAt, compliance.expiresAt),
```
mike-dydx: @rosepuppy based on when `didSetComplianceStatus` is called, it seems like we should be passing in the entire compliance object instead of just the non-geo fields. Double checking I am not missing something here though.

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
mike-dydx: ah yea, this is in a testing-specific function, but will remove to avoid confusion 👍

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -912,7 +912,7 @@ internal open class AccountSupervisor(
             state?.transferStatuses,
             state?.restriction,
             state?.launchIncentive,
-            Compliance(state?.compliance?.geo, compliance.status, compliance.updatedAt),
+            compliance,
```
rosepuppy, line 915: Wait no, it's like this because geo is actually not set in AccountSupervisor, but instead set here:
https://github.com/dydxprotocol/v4-abacus/blob/main/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/StateManagerAdaptorV2.kt#L674

So if we just set compliance like this, geo would get over written to null.

but agree the code is not great, broken up like this. There is probably a better way to do it 🤔

rosepuppy (approved):

PR #370, opened 2024-05-21
v1.7.38: Handle channel_batch_data for v4_parent_subaccount

jaredvu:
2 fixes included
- `channel_batch_data` for orders when subscribed to `v4_parent_subaccount` has `subaccountNumber` at the top-level of the contents and not within the list of `orders` object. Re-introduce `subaccountNumber` to the Order object in Abacus.
- [TRA-253](https://linear.app/dydx/issue/TRA-253/abacus-hide-leverage-on-market-order) Set `needsLeverage` to false when `TradeInput` is set to `Market + Isolated`
- Add isolatedMargin short-term order transaction test


src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -846,6 +846,18 @@ data class SubaccountFillResources(
...
+@Serializable
+enum class MarginMode(val rawValue: String) {
+    ISOLATED("ISOLATED"),
+    CROSS("CROSS");
+
+    companion object {
+        operator fun invoke(rawValue: String) =
+            MarginMode.values().firstOrNull { it.rawValue == rawValue }
+    }
+}
```
jaredvu, line 859: @mike-dydx MarginMode exists in `TradeInput.kt`, should we remove it there and opt to use the `Account.kt` one everywhere? We should also wait to rebase over @johnqh 's changes

src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -410,7 +412,8 @@ data class SubaccountPosition(
                         existing.freeCollateral !== freeCollateral ||
                         existing.marginUsage !== marginUsage ||
                         existing.quoteBalance !== quoteBalance ||
-                        existing.equity !== equity
+                        existing.equity !== equity ||
+                        existing.marginMode !== marginMode
```
prashanDYDX, line 416: nit: marginMode is a value not an object, so can use != here

src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -662,6 +666,8 @@ data class SubaccountOrder(
     val reduceOnly: Boolean,
     val cancelReason: String?,
     val resources: SubaccountOrderResources,
+    val subaccountNumber: Int?,
```
prashanDYDX, line 669: shouldn't we always have a subaccount number? (default to 0?)

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AccountProcessor.kt
```diff
@@ -299,13 +299,14 @@ internal open class SubaccountProcessor(parser: ParserProtocol) : BaseProcessor(
         modified = transform(modified, payload, "current", currentAccountKeyMap)
 
         if (firstTime) {
+            val subaccountNumber = parser.asInt(modified["subaccountNumber"])
```
prashanDYDX, line 302: `parser.asInt` returns a nullable value, should we default to 0 here? if so, may want to make a helper function for this since we do this a lot in this PR.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/OrderProcessor.kt
```diff
@@ -202,6 +204,11 @@ internal class OrderProcessor(parser: ParserProtocol) : BaseProcessor(parser) {
             if (modified["id"] == null) {
                 modified.safeSet("id", payload["clientId"])
             }
+            parser.asInt(modified["subaccountNumber"])?.run {
+                modified.safeSet("subaccountNumber", this)
```
prashanDYDX, line 208: personal preference, feel free to disregard if you disagree: i'm not a huge fan of using `.run { }` to pass `this` into things, since at a glance it can initially be ambiguous what `this` is.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/OrdersProcessor.kt
```diff
@@ -12,13 +13,20 @@ internal class OrdersProcessor(parser: ParserProtocol) : BaseProcessor(parser) {
     internal fun received(
         existing: Map<String, Any>?,
         payload: List<Any>?,
-        height: BlockAndTime?
+        height: BlockAndTime?,
+        subaccountNumber: Int?,
```
prashanDYDX, line 17: are we ever actually passing in a null subaccountNumber here?

prashanDYDX (approved):

build.gradle.kts
```diff
@@ -51,7 +51,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.7.31"
+version = "'1.7.31"
```
jaredvu, line 54: @mike-dydx is this from the auto increment script or is this a remnant of our edits

src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
jaredvu, line 669: just want to double check that this won't cause issues due to changing param order? If we are changing order here, I think we need to touchup `SubaccountOrder` in docs as well.

v4_abacus.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'v4_abacus'
-    spec.version                  = '1.7.31'
+    spec.version                  = ''1.7.31'
```
jaredvu, line 3: hmm seems like an issue with the auto increment

mike-dydx: i probably fat fingered it when entering the value, good catch, updated d417bc77b118fe256edf130bd88679a1e50fa2b2

src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
mike-dydx: not worth it then! good thinking
d417bc77b118fe256edf130bd88679a1e50fa2b2

src/commonTest/kotlin/exchange.dydx.abacus/payload/v4/V4ParentSubaccountTests.kt
```diff
@@ -329,7 +339,7 @@ class V4ParentSubaccountTests : V4BaseTests(true) {
                         "trade": {
                             "marginMode": "CROSS",
                             "options": {
-                                "needsMarginMode": false
+                                "needsMarginMode": true
```
mike-dydx: @jaredvu do we have a test for when we expect `needsMarginMode` to be false?

mike-dydx (approved):

PR #371, opened 2024-05-21
add cumulative amount to historical trading rewards + fix pagination

moo-onthelawn:
For our trading rewards chart, we want to show the cumulative rewards (not just the rewards earned from start date to end date) over time, so adding a `cumulativeAmount` field to `HistoricalTradingReward`. 
- Since indexer only returns the rewards per period, we manually calculate this in abacus in `createHistoricalTradingRewards` based off of `TradingRewards.total` 
  - at every date period, the `cumulativeAmount` is just `total - amount` (and then you decrement the running total).
- Wrote `TradingRewardsTests` to verify correct calculation; it's currently only written for the `DAILY` case (because that's the only one that'll be used in updated designs)

Other changes:
- Changed the historical trading rewards period default to be `DAILY` (in line with updated [designs](https://www.figma.com/design/dhSd8kkcUrY3UBRsdS89kH/dYdX-%E2%80%BA-Legal?node-id=0-1&t=AxhKRy2LkxMKS73O-0))
- Fixed [pagination bug](https://github.com/dydxprotocol/v4-abacus/pull/371/files#diff-a86beb396f4bee8054a9609a411467cc5fc6485bcdfe7e366e26c5081bcfe2a3L1413) in loading trading rewards from indexer where we were only ever getting at most 100 results (because one indexer batched request returns <= 100 results); now we make multiple requests to fetch all available results (and just add them to our existing results)

## Testing
- Built locally and verified that we are now fetching >100 results when applicable, verified headers in network call
- Built locally and verified values look correct on chart

Screenshot 2024-05-22 at 11 32 00 AM


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/StateManagerAdaptor.kt
```diff
@@ -1410,9 +1410,9 @@ open class StateManagerAdaptor(
             url,
             historicalTradingRewardsInPeriod,
             "startedAt",
-            0.days,
-            180.days,
-            "endedAt",
+            1.days,
+            180.days, // OTE-337: update this to read in the trading rewards start date from env.json
```
moo-onthelawn, line 1414: this works for now (because earliest date should be Jan 10/29 depending on whether we're in test or mainnet), but should be calculated from the env.json start date which is still in progress. Will update if that makes it in before this makes it in, otherwise will make new PR

rosepuppy (approved):
LGTM!

PR #375, opened 2024-05-22
Feature/tra 259 needs margin mode

johnqh:
Derived TradeInput.options.needsMarginMode so FE can use it for UX logic, whether to allow user to switch margin mode.

If it is set to false, then FE should take whichever margin mode set by Abacus. If it is set to true, give user the choice.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
@@ -1182,6 +1182,10 @@ internal class TradeInputCalculator(
                         }
 
                         "marginMode" -> {
+                            options.safeSet(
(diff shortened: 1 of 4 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
```diff
@@ -867,76 +890,59 @@ internal class TradeInputCalculator(
...
-        reduceOnly: Map<String, Any>?,
-    ): List<Map<String, Any>> {
-        return if (reduceOnly != null) {
-            val modified = list.toMutableList()
-            modified.add(reduceOnly)
-            modified
-        } else {
-            list
-        }
-    }
```
jaredvu, line 938: was this entire function just to guard for `reduceOnly`?

PR #376, opened 2024-05-22
update trading rewards to fetch all rewards since rewards_history_start_date defined in env.json

moo-onthelawn:
- Update the `Environment` type in abacus with `rewardsHistoryStartDateMs` from [here](https://github.com/dydxprotocol/v4-web/pull/578/files#diff-8817bf86fd45bfdf9c54b3c3c52f6e944fb4a05bc7d9f3b9cfd448fc8932130eR299)
- Remove hardcoded `180 days` from query for trading rewards, and instead calculate based on `now - startDateMs + 2.days` (for buffer). Default to 180 days from start if for some reason env.json isn't configured to have this
- Fix previous placeholder

## Testing
- Built locally, verified that data is fetched all the way to start date, tested with a non-null and null env.json variable

Screenshot 2024-05-22 at 4 34 29 PM

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/Enviroment.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/Enviroment.kt
@@ -4,9 +4,11 @@ import exchange.dydx.abacus.protocols.LocalizerProtocol
 import exchange.dydx.abacus.protocols.ParserProtocol
 import exchange.dydx.abacus.utils.IList
 import exchange.dydx.abacus.utils.IMap
+import exchange.dydx.abacus.utils.ServerTime
 import kollections.JsExport
 import kollections.iMutableMapOf
 import kollections.toIList
+import kotlin.time.Duration.Companion.days
 
 @JsExport
 data class IndexerURIs(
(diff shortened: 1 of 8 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/Enviroment.kt
```diff
@@ -455,6 +458,7 @@ class V4Environment(
             val squidIntegratorId = parser.asString(data["squidIntegratorId"])
             val chainName = parser.asString(data["chainName"])
             val chainLogo = parser.asString(data["chainLogo"])
+            val rewardsHistoryStartDateMs = parser.asString(data["rewardsHistoryStartDateMs"]) ?: return null
```
yogurtandjam, line 461: [q] do we want to early return null here? maybe it's worth just having a default value (180 days maybe?)

moo-onthelawn: oh good point 🤔 ideally it should never be null, but probably not worth breaking/early returning on it - will make change

PR #381, opened 2024-05-23
Fix publishToMavenLocal

prashanDYDX:
need to create expect/actual for kotlin-inject component creation methods for every final platform. (See kotlin-inject KMP examples.)

https://github.com/evant/kotlin-inject-samples/blob/main/multiplatform/echo/src/macosArm64Main/kotlin/App.macosArm64.kt


ruixhuang (approved):
It's a lot of boilerplate code.

prashanDYDX:
@ruixhuang yeah, it's because KSP support for Kotlin Multiplatform isn't very good. Luckily we shouldn't need to touch this much

PR #388, opened 2024-05-28
Add equity tiers in configs output

aforaleka:
need to render an equity tier table in FE similar to this, with tier 1..6
Screenshot 2024-05-28 at 11 48 58 AM

we're also removing equity tier limit validation on the FE, which will be done in another PR. therefore ignoring any duplicated processing logic here

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/output/Configs.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/output/Configs.kt
@@ -252,8 +252,8 @@ data class FeeTier(
 @JsExport
 @Serializable
 data class EquityTiers(
-    val shortTermOrderEquityTiers: List<EquityTier>,
-    val statefulOrderEquityTiers: List<EquityTier>,
+    val shortTermOrderEquityTiers: IList<EquityTier>,
+    val statefulOrderEquityTiers: IList<EquityTier>,
 ) {
     companion object {
         internal fun create(
@@ -297,6 +297,7 @@ data class EquityTiers(
                 }
                 return equityTiers
             }
+            Logger.d { "Equity Tiers values not valid" }
             return null
         }
     }
--- a/src/commonMain/kotlin/exchange.dydx.abacus/output/Configs.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/output/Configs.kt
@@ -6,7 +6,9 @@ import exchange.dydx.abacus.protocols.ParserProtocol
 import exchange.dydx.abacus.utils.IList
 import exchange.dydx.abacus.utils.Logger
 import kollections.JsExport
+import kollections.iListOf
 import kollections.iMutableListOf
+import kollections.toIList
 import kotlinx.serialization.Serializable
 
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/output/Configs.kt
```diff
@@ -247,6 +249,105 @@ data class FeeTier(
     }
 }
 
+@JsExport
+@Serializable
+data class EquityTiers(
+    val shortTermOrderEquityTiers: List<EquityTier>,
```
moo-onthelawn, line 255: hmm should this more accurately be `IList<EquityTier>`?

src/commonMain/kotlin/exchange.dydx.abacus/output/Configs.kt
```diff
@@ -247,6 +249,105 @@ data class FeeTier(
...
+                    parser.asMap(item)?.let {
+                        val tier = existing?.getOrNull(i)
+                        val nextTierData = parser.asMap(nextItem)
+                        EquityTier.create(tier, parser, it, i + 1, nextTierData)?.let { equityTier ->
+                            equityTiers.add(equityTier)
+                        }
+                    }
+                }
+                return equityTiers
+            }
```
moo-onthelawn, line 299: nit: should we also add an invalid log here

moo-onthelawn (commented):
can we also replace the `Map<String, Any>?` typings in the codebase (i.e.  in `EquityTiersProcessor`) and the existing `EquityTier` type in `TradeInputDataValidator` (otherwise it gets a lil confusing to read)?

src/commonMain/kotlin/exchange.dydx.abacus/output/Configs.kt
aforaleka: mmm it seems like they're aliases so IList = List? but i'll change it!

v4_abacus.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'v4_abacus'
-    spec.version                  = '1.7.32'
+    spec.version                  = '1.7.33'
```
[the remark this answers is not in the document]
aforaleka: yeee i'll rebase last minute before merge so i know which version to bump to

src/commonMain/kotlin/exchange.dydx.abacus/output/Configs.kt
```diff
@@ -247,6 +249,106 @@ data class FeeTier(
...
+            return null
+        }
+
+        internal fun create(
+            existing: IList<EquityTier>?,
+            parser: ParserProtocol,
+            data: List<*>?
+        ): IList<EquityTier>? {
+            data?.let {
+                val equityTiers = iMutableListOf<EquityTier>()
```
yogurtandjam, line 286: [nit]: 'creating {object name}' logs

yogurtandjam (approved):
LGTM

PR #394, opened 2024-05-29
feat: skip v1/assets endpoint [OTE-346]

yogurtandjam:
implements skip v1/assets endpoint into abacus

TODO: write a real description


src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
```diff
@@ -56,7 +60,24 @@ internal class SkipProcessor(
         existing: Map<String, Any>?,
         payload: Map<String, Any>
     ): Map<String, Any>? {
-        throw NotImplementedError("receivedTokens is not implemented in SkipProcessor!")
+        if (this.chains != null && this.skipTokens != null) {
+            return existing
+        }
+
+        val chainToAssetsMap = payload["chain_to_assets_map"] as Map<String, Map<String, List<Map<String, Any>>>>?
```
rosepuppy, line 67: question: Does it make senes to use parser here?

rosepuppy (approved):

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
yogurtandjam: the parser unfortunately doesn't have the ability to type nested primitives 😢 

we could parseAsMap/parseAsNativeMap but then it defaults it to `Map<String, Any>`.
IMO the best path forward is to create structs for pretty much everything so we don't have to deal with maps anymore, but i think that's a bit of a ways off

PR #396, opened 2024-05-29
chore: remove squidV1 codepaths

yogurtandjam:
https://dydx-team.slack.com/archives/C06FR0QLWFJ/p1716998653564039

we don't use squidV1. removing this codepath to make prep work for squid -> skip migration simpler


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor+Transfer.kt
```diff
@@ -73,7 +73,7 @@ internal fun V4StateManagerAdaptor.retrieveDepositExchanges() {
 internal fun V4StateManagerAdaptor.retrieveDepositRoute(state: PerpetualState?) {
     val isCctp = state?.input?.transfer?.isCctp ?: false
     when (appConfigs.squidVersion) {
-        AppConfigs.SquidVersion.V1, AppConfigs.SquidVersion.V2WithdrawalOnly -> retrieveDepositRouteV1(
```
prashanDYDX, line 76: can we just remove the V1 enum value instead? it's a breaking change, but we've already confirmed it to be safe

ruixhuang (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor+Transfer.kt
yogurtandjam: im confused - we are removing the [enum value](https://github.com/dydxprotocol/v4-abacus/pull/396/files#diff-327496ce92a886e88a5f7bac649f6e1570ea3b684e1fb788f10657ffa5069383L13)

prashanDYDX: oh oops i'm blind disregard

PR #399, opened 2024-05-30
Ensure all Json.parseToJsonElement calls are wrapped in try-catch wit…

prashanDYDX:
…h appropriately scoped exceptions.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -51,7 +51,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.7.60"
+version = "1.7.61"
 
 repositories {
     google()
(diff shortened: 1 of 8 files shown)


yogurtandjam (approved):
feels like we may want to make a `jsonSafe` helper method and (if possible?) make a linting rule to not allow usage/import of the `json.Json` library if we want json to always be a safe operation.

that being said - are there times when the app *should* fail and throw if we can't jsonify a payload?

prashanDYDX:
@yogurtandjam yes good callouts, i shall tackle in a follow up

prashanDYDX:
@yogurtandjam re failing on payloads, likely yes - but i don't have enough context on these existing usages to say which ones should throw, so just made them safe for now.

PR #401, opened 2024-05-30
create memo field for transferInput

moo-onthelawn:
- filled out `TransferInput.md` a bit more (it's outdated)
- add `memo` to `TransferInput`; clear it when transfer type switches (but maintain it between token type switches)

- I considered adding a `needsMemo` field to the different transfer type options (`WithdrawalInputOptions`, etc) but that seemed like overkill when the caller can just conditionally render the memo field depending on type - and maybe we'd want to allow users to fill in a memo in other cases in the future?

Tested with tests; built locally

https://github.com/dydxprotocol/v4-abacus/assets/70078372/73acda34-c595-46ef-8e9e-db4702d20b7a

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+TransferInput.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+TransferInput.kt
@@ -25,6 +25,7 @@ enum class TransferInputField(val rawValue: String) {
     chain("chain"),
     token("token"),
     address("address"),
+    MEMO("memo"),
     fastSpeed("fastSpeed");
 
     companion object {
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+TransferInput.kt
```diff
@@ -25,6 +25,7 @@ enum class TransferInputField(val rawValue: String) {
     chain("chain"),
     token("token"),
     address("address"),
+    MEMO("memo"),
```
moo-onthelawn, line 28: lol this is for linting purposes, but yeah the other enum vals should be renamed at some point (would rather not incorporate into this PR)

rosepuppy (approved):

PR #406, opened 2024-05-31
Move the file read/write to IO thread

ruixhuang:
Should impact mobile only..  Tested on both iOS and Android


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
```diff
@@ -282,27 +282,35 @@ class AsyncAbacusStateManagerV2(
...
+            }
         }
     }
 
     private fun loadFromRemoteConfigFile(configFile: ConfigFile) {
         ioImplementations.fileSystem?.readCachedTextFile(
             configFile.path,
         )?.let {
-            parse(it, configFile)
+            ioImplementations.threading?.async(ThreadingType.abacus) {
```
prashanDYDX, line 311: why `abacus` for this one and the one below?

prashanDYDX, line 311: i feel like this is a good example of how threading is making things hard to reason about. when we call `loadFromRemoteConfigFile`, we are actually on the main thread since this is called at object initialization time. So shouldn't we be going back to that thread instead of the `abacus` thread?

prashanDYDX (approved):
this change seems fine, but as i mention in my comment i think highlights the issues with how we handle threading in abacus.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
ruixhuang: Internal processing is done on the abacus thread.  But I missed the post-processing after the rest call.

Making readCachedTextFile main-safe would require a breaking changing to the protocol to use callbacks.  For now I think we can live with what we have.  Yes, it does have some threading complexity in Abacus, but the touchpoints are rather limited.  Once we have those working properly, it shouldn't require much future maintenance.

PR #411, opened 2024-06-03
Deprecate FOK

aforaleka:
deprecating FOK since protocol is removing FOK in v6
- abacus will remove FOK as a timeInForce / execution option
- note: there's one FOK mention left in abacus as localized string of the `timeInForceStringKey`, I'm leaving it in as we still want to show that as order details if trader submits it via API and checks it on the FE


prashanDYDX (approved):
could maybe also use this opportunity to move checks on `"IOC"` string to the enum, but up to your judgment based on effort

PR #412, opened 2024-06-04
Add balance processing for unbonding and staking rewards

rosepuppy:


src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AccountProcessor.kt
```diff
@@ -952,6 +957,56 @@ internal class V4AccountProcessor(parser: ParserProtocol) : BaseProcessor(parser
...
+
+    internal fun receivedStakingRewards(
+        existing: Map<String, Any>?,
+        payload: Map<String, Any>?,
+    ): Map<String, Any>? {
+        val rewards = parser.asList(parser.value(payload, "rewards"))
+        val total = parser.asList(parser.value(payload, "total"))
+        if (payload == null || rewards == null || total == null) {
+            return existing
+        }
```
moo-onthelawn, line 993: hmm do we need to be checking for `rewards == null || total == null` here? (Sorta feels like the for loops will just be skipped and we'll correctly set stakingRewards to null?

moo-onthelawn (approved):
lgtm with one small q!

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AccountProcessor.kt
rosepuppy: Mainly did it due to typing issue, but I think its ok to return early if these are null

PR #415, opened 2024-06-04
feat: top level deposit methods [OTE-388]

yogurtandjam:
adds top level deposit methods.

TODO: tests


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/V4StateManagerAdaptor+Transfer.kt
```diff
@@ -212,67 +207,6 @@ private fun V4StateManagerAdaptor.retrieveDepositRouteV2(state: PerpetualState?)
     }
 }
 
-private fun V4StateManagerAdaptor.retrieveSkipDepositRoute(state: PerpetualState?) {
```
yogurtandjam, line 215: i think this is a weird git thing. this code isn't even in the previous branch.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -188,6 +188,139 @@ internal class OnboardingSupervisor(
...
+    @Suppress("UnusedPrivateMember", "ForbiddenComment")
+    private fun retrieveSkipDepositRouteCCTP(
+        state: PerpetualState?,
+        accountAddress: String,
+        sourceAddress: String,
+        subaccountNumber: Int?,
+    ) {
+//      We have a lot of duplicate code for these deposit/withdrawal route calls
+//      It's easier to dedupe now that he url is the same and only the args differ
+//      TODO: Consider creating generateArgs fun to reduce code duplication
```
rosepuppy, line 265: nice TODO!

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -188,6 +188,139 @@ internal class OnboardingSupervisor(
...
+            val header = iMapOf(
+                "Content-Type" to "application/json",
+            )
+            helper.post(url, header, body.toJsonPrettyPrint()) { _, response, code, headers ->
+                if (response != null) {
+                    val currentFromAmount = stateMachine.state?.input?.transfer?.size?.size
+                    val oldFromAmount = oldState?.input?.transfer?.size?.size
+                    val requestId = helper.parser.asString(headers?.get("x-request-id"))
+                    if (currentFromAmount == oldFromAmount) {
+                        update(stateMachine.squidRoute(response, subaccountNumber ?: 0, requestId), oldState)
```
rosepuppy, line 247: nit: Assuming you are going to rename squidRoute in the future

rosepuppy (approved):

rosepuppy (commented):
Looks like there is build failure 🧐

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
yogurtandjam: yup - i can add a TODO for this as well - was hesitant to rename right now because the new skip endpoint isn't even `route` so their names are a bit incompatible.

yogurtandjam:
> Looks like there is build failure 🧐

ah yeahhh - i did a rebase fail and lost an import 😢 - fix is coming in a sec

PR #416, opened 2024-06-05
feat: skip route + msgs endpoints [OTE-348]

yogurtandjam:
implements skip replacement for squid /route endpoint

This PR implements the skip analog of squid's /route endpoint by utilizing Skip's msgs_direct endpoint.

In order to do this we:

- Implemented the receivedRoute method on the SkipProcessor
- Implemented the receivedRouteV2 method on the SkipProcessor
- Created and implemented the `SkipRouteProcesor
- Created and implemented the SkipRoutePayloadProcessor
- Skip does not provide all the same data that Squid does, so we did our best to keep the data format as similar as possible. We'll have to make some updates on the web side to handle these changes though. Most notably:
  - gasLimit should no longer be required

Some things to cleanup after this PR:
- errors property is set in a pretty hacky way to coerce the skip payload to mimic squid's. this is intentional to reduce the # of breaking internal API changes. We can migrate that after we confirm core functionality works.
- slippage is just hard coded. we made sure that the input param for the skip payload is synced with the processor but we 
should be updating this value all across the codebase. this will be a separate PR as it's not directly related to this functionality
- there's still some shared logic we may want to deduplicate from squid and skip (TBD. since we're removing squid it may not be worth it)

Future architectural things:
- i don't think we should be imputing the to and from addresses. we'd be able to easily dedupe a lot more code if we depended on the users dispatching the toAddress toTokenDenom toChainId transferType etc. in the transfer request.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
@@ -12,7 +12,7 @@ import exchange.dydx.abacus.state.manager.CctpConfig.cctpChainIds
 import exchange.dydx.abacus.utils.mutable
 import exchange.dydx.abacus.utils.safeSet
 
-@Suppress("NotImplementedDeclaration")
+@Suppress("NotImplementedDeclaration", "ForbiddenComment")
 internal class SkipProcessor(
     parser: ParserProtocol,
     private val internalState: InternalTransferInputState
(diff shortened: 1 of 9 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
```diff
@@ -85,19 +85,45 @@ internal class SkipProcessor(
...
+        val tokenAddress = parser.asString(parser.value(modified, "transfer.token"))
+        val selectedChainId = parser.asString(parser.value(modified, "transfer.chain"))
+        val decimals = parser.asDouble(selectedTokenDecimals(tokenAddress = tokenAddress, selectedChainId = selectedChainId))
+        val processor = SkipRouteProcessor(parser)
+
+        modified.safeSet(
+            "transfer.route",
+            processor.received(null, payload, decimals = decimals) as MutableMap<String, Any>,
+        )
+        if (requestId != null) {
```
rosepuppy, line 101: Does skip api have requestId concept as well?

rosepuppy (approved):

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
yogurtandjam: they do send back a `Grpc-Metadata-Traceid` which we can use as a unique identifier for the request. 

that being said - i don't think we need that param anymore. i think it was only used in the `fetchSquidStatus` method in web as a header. the skip API won't require us to store a trace/request id

yogurtandjam: ah i just realized, there was an opportunity to clean this up. will do in a future PR 🫡

PR #417, opened 2024-06-05
feat: top level withdrawal methods [OTE-389]

yogurtandjam:
followup to: https://github.com/dydxprotocol/v4-abacus/pull/415
adds top level withdrawal methods.
retrieveSkipWithdrawalRouteNonCCTP matches retrieveWithdrawalRouteV1
retrieveSkipWithdrawalRouteCCTP matches retrieveWithdrawalRouteV2
retrieveSkipWithdrawalRouteExchange matches retrieveWithdrawalRouteNoble

NOTE: to test this properly i think we'll need mockk or some other library to mock our API calls or some mock helper class that we can inject with API responses. I figure it's worth having a small discussion about this. In the meantime these methods are not used so it should be safe to merge as is. Come closer to QA time I can spend time writing extensive testing


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -884,6 +884,176 @@ internal class OnboardingSupervisor(
...
+        val url = helper.configs.skipV2MsgsDirect()
+        val fromAddress = accountAddress
+        if (
+            fromAmount != null &&
+            fromAmount > 0 &&
+            fromAmountString != null &&
+            chainId != null &&
+            dydxTokenDemon != null &&
+            squidIntegratorId != null
+        ) {
```
ruixhuang, line 915: Nit: Can we add an `else` to call `Logger.e { }` to log an error?   Same with other functions.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -884,6 +884,176 @@ internal class OnboardingSupervisor(
...
+    ) {
+        val usdcSize = helper.parser.asDecimal(state?.input?.transfer?.size?.usdcSize)
+        val fromAmount = if (usdcSize != null && usdcSize > gas) {
+            ((usdcSize - gas) * Numeric.decimal.TEN.pow(decimals)).toBigInteger()
+        } else {
+            null
+        }
+        val chainId = helper.environment.dydxChainId
+        val squidIntegratorId = helper.environment.squidIntegratorId
+        val dydxTokenDemon = helper.environment.tokens["usdc"]?.denom
```
ruixhuang, line 904: Can we use nativeTokenDemon instead?  Legal prefers "native" over "dydx" in our codebase.

Also, tokens["usdc"] would be the UDSC instead Native token, so the variable name should change accordingly.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -884,6 +884,176 @@ internal class OnboardingSupervisor(
...
+                ),
+                "slippage_tolerance_percent" to SLIPPAGE_PERCENT,
+            )
+
+            val oldState = stateMachine.state
+            val header = iMapOf(
+                "Content-Type" to "application/json",
+            )
+            helper.post(url, header, body.toJsonPrettyPrint()) { _, response, code, headers ->
+                if (response != null) {
```
ruixhuang, line 934: Might want to check the http code as well.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
yogurtandjam: ah got it thanks! this was c/p'ed from another method and was confused by this naming as well. thanks :)

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
yogurtandjam: hmm wait do you mean if we don't have all the args? are we sure that's what we want? 

any transfer state change will fire off a retrieve deposit/withdrawal route function even if it doesn't make sense to do so, like when we clear the modal after submitting a withdraw/deposit. If we add an error log we'll be logging an error for all of those instances

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
yogurtandjam: we actually accept 4XX codes and throw the error msg into state so the users can see why their withdraw/deposit attempts are failing!

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
yogurtandjam: renamed to `nativeChainUSDCDenom` since `nativeToken` isn't quite accurate.

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+Squid.kt
```diff
@@ -39,6 +39,8 @@ internal fun TradingStateMachine.squidV2SdkInfo(payload: String): StateChanges?
     }
 }
 
+@Suppress("ForbiddenComment")
```
prashanDYDX, line 42: nit you can make a ticket and use the ticket number instead of TODO

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
prashanDYDX: perhaps we should just `error()` out for things like `fromChain` and `nativeChainUSDCDenom` being null.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -884,6 +937,187 @@ internal class OnboardingSupervisor(
...
+                    toChain to toAddress,
+                ),
+                "slippage_tolerance_percent" to SLIPPAGE_PERCENT,
+            )
+
+            val oldState = stateMachine.state
+            val header = iMapOf(
+                "Content-Type" to "application/json",
+            )
+            helper.post(url, header, body.toJsonPrettyPrint()) { _, response, code, headers ->
```
prashanDYDX, line 988: why `.toJsonPrettyPrint()` ?

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -884,6 +937,187 @@ internal class OnboardingSupervisor(
...
+        decimals: Int,
+        gas: BigDecimal,
+        accountAddress: String,
+        sourceAddress: String,
+        subaccountNumber: Int?,
+    ) {
+        val toChain = state?.input?.transfer?.chain
+        val toToken = state?.input?.transfer?.token
+        val toAddress = state?.input?.transfer?.address
+        val usdcSize = helper.parser.asDecimal(state?.input?.transfer?.size?.usdcSize)
```
prashanDYDX, line 1009: for any of these that are required, can we just return early?
`val toChain = state?.input?.transfer?.chain ?: return`

same for other methods in this pr

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -1236,4 +1506,110 @@ internal class OnboardingSupervisor(
...
+        state: PerpetualState?,
+        decimals: Int,
+        gas: BigDecimal,
+        accountAddress: String,
+        subaccountNumber: Int?,
+        callback: TransactionCallback
+    ) {
+//      We have a lot of duplicate code for these deposit/withdrawal route calls
+//      It's easier to dedupe now that the url is the same and only the args differ
+//      TODO: Consider creating generateArgs fun to reduce code duplication
```
prashanDYDX, line 1521: nit please make tickets instead of leaving TODOs

prashanDYDX: if this is something you are going to tackle as a follow up, can just leave a comment in the pr

ruixhuang (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
yogurtandjam: post body arg accepts a string - just following the convention we've been using thus far in other fns for str formatting

PR #420, opened 2024-06-07
isolated margin: use correct payload + codespace broadcast errors

aforaleka:
- broadcast errors from different module/codespace have different meanings; adding codespace to localized stringkey to differentiate
- use order payload instead of transfer in the place order / isolated margin flow, so that client has access to order client Id when there's an error

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
@@ -685,7 +685,9 @@ internal class SubaccountSupervisor(
                     useTransactionQueue,
                 )
             } else {
-                helper.send(error, callback, transferPayload)
+                // callback with order payload instead of transfer payload since
+                // client shows it as a place order error and needs order client id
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -698,7 +700,7 @@ internal class SubaccountSupervisor(
                 transferPayloadString,
                 null,
                 isolatedMarginTransactionCallback,
-                useTransactionQueue,
+                true,
```
prashanDYDX, line 703: let's use named params for booleans

prashanDYDX (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
ruixhuang: We should try to use named params for all calls if possible.  

Android Studio will do this for you.  Click on the light-bulb will give you the option:

Screenshot 2024-06-06 at 18 46 12

PR #428, opened 2024-06-09
Fix to Position not displaying after fill

jaredvu:
Null position should be removed as it overwrites the update to `perpetualPosition` from the WS

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginModeCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginModeCalculator.kt
@@ -2,7 +2,6 @@ package exchange.dydx.abacus.calculator
 
 import abs
 import exchange.dydx.abacus.protocols.ParserProtocol
-import exchange.dydx.abacus.utils.Logger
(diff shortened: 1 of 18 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginModeCalculator.kt
```diff
@@ -126,48 +131,54 @@ internal object MarginModeCalculator {
...
-                if (containsMarket(parser, subaccount, marketId)) {
-                    return childSubaccountNumber
-                } else {
-                    if (childSubaccountNumber > lastSubaccountNumber) {
-                        lastSubaccountNumber = childSubaccountNumber
-                    }
-                }
+        val subaccounts = parser.asNativeMap(account?.get("subaccounts")) ?: return subaccountNumber
+
+        val utilizedSubaccountsMarketIdMap = subaccounts.mapValues {
```
prashanDYDX, line 135: i've seen this calculated in a few places, should we just expose this as a helper function?

prashanDYDX (approved):

PR #433, opened 2024-06-11
CT-917: Map liquidation order type to limit.

prashanDYDX:


src/commonMain/kotlin/exchange.dydx.abacus/output/input/TradeInput.kt
```diff
@@ -678,8 +678,12 @@ enum class OrderType(val rawValue: String) {
     ;
 
     companion object {
-        operator fun invoke(rawValue: String?) =
-            OrderType.values().firstOrNull { it.rawValue == rawValue }
+        operator fun invoke(rawValue: String?): OrderType? {
+            // hacky fix until indexer returns underlying type
```
moo-onthelawn, line 682: nit: can we link CT-916 here

src/commonMain/kotlin/exchange.dydx.abacus/output/input/TradeInput.kt
```diff
@@ -678,8 +678,12 @@ enum class OrderType(val rawValue: String) {
...
+            if (rawValue == liquidation.rawValue) return limit
+            return OrderType.values().firstOrNull { it.rawValue == rawValue }
```
moo-onthelawn, line 684: I'm bad at kotlin but do we not have to return `OrderType.limit`? (instead of just `limit`?

moo-onthelawn (approved):
if it's easy, can we add a test?

src/commonMain/kotlin/exchange.dydx.abacus/output/input/TradeInput.kt
prashanDYDX: i think because it's within the enum class itself, no we don't

moo-onthelawn (approved):
thank u for the testsss

PR #438, opened 2024-06-11
feat: create trading rewards by period with zeros removed

tyleroooo:
first abacus pr, not really sure what's going on in here


src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -1791,26 +1792,32 @@ data class TradingRewards(
...
                 val total = parser.asDouble(data["total"])
-                val historical = total?.let {
+                val fullHistory = total?.let {
                     createHistoricalTradingRewards(
                         it,
-                        existing?.historical,
-                        parser.asMap(data["historical"]),
+                        existing?.fullHistory,
+                        parser.asMap(data["fullHistory"]),
+                        true,
```
prashanDYDX, line 1800: can we use named params for booleans? (ideally all params) same with the other call below

tyleroooo: sure

build.gradle.kts
```diff
@@ -51,7 +51,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.7.71"
+version = "1.7.72"
```
prashanDYDX, line 54: let's manually bump minor version instead, we made a breaking API change here (TradingRewards is consumed by clients directly)

src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -1780,7 +1780,8 @@ internal fun Instant.previousMonth(): Instant {
 data class TradingRewards(
     val total: Double?,
     val blockRewards: IList<BlockReward>?,
-    val historical: IMap<String, IList<HistoricalTradingReward>>?
+    val fullHistory: IMap<String, IList<HistoricalTradingReward>>?,
```
prashanDYDX, line 1783: wdyt about `filledHistory` and `rawHistory`?

prashanDYDX:
can we add some tests for the eventHistory field?

src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -1938,15 +1957,21 @@ data class TradingRewards(
...
-                        result.add(synced)
+                        if (fillZeros) {
+                            addHistoricalTradingRewards(result, synced!!, period, lastStart)
+                        }
+                        result.add(synced!!)
                         dataIndex++
                         lastStart = synced.startedAtInMilliseconds
                         cumulativeAmount = cumulativeAmount - synced.amount
+                    } else {
+                        dataIndex++
```
tyleroooo, line 1968: This seemed like an infinite loop technically

src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -2019,7 +2044,7 @@ data class TradingRewards(
                 Instant.fromEpochMilliseconds(lastStart.toLong())
             }
             while (obj.startedAtInMilliseconds < lastStartTime.toEpochMilliseconds().toDouble()) {
-                val previous = previousPlaceHolder(period, lastStartTime, obj.cumulativeAmount + obj.amount)
```
tyleroooo, line 2022: this feels like a bug and when I made my tests it really seems like it shouldn't add amount here.

tyleroooo: These objects with zero amount shouldn't be adding to the older cumulative.

tyleroooo:
Haven't tested locally yet, will do tomorrow

tyleroooo:
Tested, seems good

tyleroooo:
@prashanDYDX are there outstanding action items for me here?

build.gradle.kts
```diff
@@ -51,7 +51,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.7.74"
+version = "1.8.0"
```
ruixhuang, line 54: The convention is to update the minor version number when there is a breaking change (meaning the client code will need to be updated after bumping to this Abacus version).

In your case, it looks like the change is non-breaking.  Recommend just updating the patch number (or leave it to the Github Action to update the version).

ruixhuang (approved):

build.gradle.kts
ruixhuang: Oh nevermind.. a field name got change..  Yeah.. then it makes sense to update the minor version... Up to you :-)

PR #439, opened 2024-06-12
v.1.7.82: Add subaccountNumber and marginMode to fills

jaredvu:
### Description

* `output/Account`
  *  Add `subaccountNumber` and `marginMode` to `SubaccountFill` so the FE can expose margin mode.

* `AccountProcessor`
  * pass `subaccountNumber`

* `FillProcessor`
  * pass `subaccountNumber`

* `FillsProcessor`
  * derive and safe set `marginMode` from `subaccountNumber`

* `PerpetualState`
  * Return `SubaccountFills` from `childSubaccounts` as well.


src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/FillsProcessor.kt
```diff
@@ -1,15 +1,34 @@
...
 
 internal class FillsProcessor(parser: ParserProtocol) : BaseProcessor(parser) {
     private val itemProcessor = FillProcessor(parser = parser)
 
-    override fun received(existing: List<Any>?, payload: List<Any>): List<Any>? {
+    fun received(existing: List<Any>?, payload: List<Any>, subaccountNumber: Int?): List<Any>? {
         val new = payload.mapNotNull { eachPayload ->
-            parser.asNativeMap(eachPayload)?.let { eachPayloadData -> itemProcessor.received(null, eachPayloadData) }
+            parser.asNativeMap(eachPayload)?.let { eachPayloadData ->
+                val modified = eachPayloadData.toMutableMap()
```
ruixhuang, line 16: I think it'd be cleaner to do this in the FillProcessor (i.e., passing the subaccountNumber over), since technically we should modify the output instead of the input payload.

ruixhuang (approved):

PR #383, opened 2024-05-24
Disable Noble balance polling when connecting Kepler wallet

Ryz0nd:
Hello dYdX Team!

This PR addresses the ongoing issue of Noble USDC transfer tx signing requests persisting when connecting the Keplr wallet on the dYdX web platform.

We are planning to implement a separate USDC deposit modal, and we want the signing request to occur only when this modal is activated.

So, we have added the `cosmosWalletConnected` state to prevent the `pollNobleBalance` method from executing when the Kepler wallet is connected. 

If there are better approaches, we would appreciate your feedback.

Thank you.

https://github.com/dydxprotocol/v4-abacus/assets/53689198/5f1cf209-8508-4547-8e2d-fe821def8748

PR #440, opened 2024-06-13
chore: fix Autosweep Issue After StateManagerV2 Update

Ryz0nd:
I have added the missing parts to disable autosweep for Keplr wallet connection in StateManagerV2.

Related PR: #383


jaredvu:
@Ryz0nd I think most of the auto-signed transactions are triggered from the clients (v4-web, v4-native-ios, v4-native-android). We shouldn't disable balance polling within v4-abacus.

Ryz0nd:
@jaredvu Thank you for your feedback.

The link you mentioned only allows for disabling deposits/withdrawals to/from subaccounts. 
How can we disable the automatic noble balance transfers?

https://github.com/dydxprotocol/v4-abacus/blob/5d3d27d69d5ac107553ec6e5c2f9eb4b632ca27e/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt#L482
want to fix the issue of the web repeatedly requesting signatures for Noble IBC transfers.

rosepuppy:
@jaredvu I think this fix is ok. Since Noble polling and auto-sweep happens entirely in abacus.

PR #442, opened 2024-06-13
fix: orderbook calculation bugs, expose more data

tyleroooo:


src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/OrderbookProcessor.kt
```diff
@@ -510,43 +517,55 @@ internal class OrderbookProcessor(parser: ParserProtocol) : BaseProcessor(parser
     }
 
     fun group(orderbook: List<Any>?, grouping: Double): List<Any>? {
-        return if (orderbook != null && orderbook.size > 0) {
+        return if (!orderbook.isNullOrEmpty()) {
```
yogurtandjam, line 520: [discussion/suggestion] IMO this is one those functions that has enough math and logic where it makes sense to test separately. we should probably be actively testing edge cases and difficult to reason-about situations like stuff that's been commented here.

i tried to start a pattern with adding [the processor tests](https://github.com/dydxprotocol/v4-abacus/tree/main/src/commonTest/kotlin/exchange.dydx.abacus/processor) to the commonTest folder.

yogurtandjam (commented):
looks good but would really appreciate some unit tests to illustrate any edge or confusing cases we're concerned with. i'm not as familiar with this part of the code and there's a lot of logic here so i'm hesitant to stamp

src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/OrderbookProcessor.kt
```diff
@@ -510,43 +517,55 @@ internal class OrderbookProcessor(parser: ParserProtocol) : BaseProcessor(parser
...
+            // orderbook always ordered in increasing depth which is either increasing (asks) or decreasing (bids) price
+            // we want to round asks up and bids down so they don't have an overlapping group in the middle
             val firstPrice = parser.asDouble(parser.value(orderbook.firstOrNull(), "price"))!!
-            var floor = Rounder.round(firstPrice, grouping)
-            var ceiling = floor + grouping
+            val lastPrice = parser.asDouble(parser.value(orderbook.lastOrNull(), "price"))!!
```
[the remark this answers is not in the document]
tyleroooo: we don't know if these are bids or asks, right?

src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/OrderbookProcessor.kt
tyleroooo: I think the existing orderbook tests hit the main points pretty well............and I feel like it would take me forever to set up new tests.

tyleroooo: Yeah the orderbook tests basically hit this directly

yogurtandjam (approved):

PR #447, opened 2024-06-17
partially canceled order status + canceling -> pending for short term orders

aforaleka:
context: showing an order as canceled is confusing when it's partially filled, but we also want to show user that it's been canceled (so FE doesn't enable another cancel). therefore introducing a new status that FE can show text as "Partial Fill" but with a canceled icon (i.e. in Abacus we named this partiallyCanceled)

- add new partially canceled order status;
- short term order when BEST_EFFORT_CANCELED should be shown as PENDING; to support the case where an order should go from pending -> "partial fill" in the FE, instead of canceling -> "partial fill"

image


src/commonMain/kotlin/exchange.dydx.abacus/output/input/TradeInput.kt
```diff
@@ -709,7 +709,8 @@ enum class OrderStatus(val rawValue: String) {
     `open`("OPEN"),
     pending("PENDING"),
     untriggered("UNTRIGGERED"),
-    partiallyFilled("PARTIALLY_FILLED");
+    partiallyFilled("PARTIALLY_FILLED"),
+    partiallyCanceled("PARTIALLY_CANCELED");
```
moo-onthelawn, line 713: can we add a comment about the differences here? (basically your PR description)

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/OrderProcessor.kt
```diff
@@ -209,27 +209,33 @@ internal class OrderProcessor(parser: ParserProtocol) : BaseProcessor(parser) {
...
 
             modified.safeSet("cancelReason", payload["removalReason"] ?: payload["cancelReason"])
 
+            parser.asDouble(payload["orderFlags"])?.let { orderFlags ->
+                // if order is short-term order and indexer returns best effort canceled and has no partial fill
+                // treat as a pending order until it's partially filled or finalized
+                val isBestEffortCanceled = modified["status"] == "BEST_EFFORT_CANCELED"
+                val cancelReason = parser.asString(modified["cancelReason"])
+                val isUserCanceled = cancelReason == "USER_CANCELED" || cancelReason == "ORDER_REMOVAL_REASON_USER_CANCELED"
+                if (orderFlags.equals(Numeric.double.ZERO) && isBestEffortCanceled && !isUserCanceled) {
```
moo-onthelawn, line 234: nit: could we add an extra line of `isShortTermOrder = orderFlags.equals(Numeric.double.ZERO)` just to make it easier to understand

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -354,13 +354,14 @@ internal class SubaccountSupervisor(
...
                             OrderStatus.partiallyFilled -> AnalyticsEvent.TradePlaceOrderStatusPartiallyFilled
+                            OrderStatus.partiallyCanceled -> AnalyticsEvent.TradePlaceOrderStatusPartiallyCanceled
                         }
 
                         tracking(orderStatusChangeEvent.rawValue, analyticsPayload)
 
                         when (order.status) {
                             // order reaches final state, can remove / skip further tracking
-                            OrderStatus.cancelled, OrderStatus.filled -> {
+                            OrderStatus.cancelled, OrderStatus.partiallyCanceled, OrderStatus.filled -> {
```
moo-onthelawn, line 364: I wonder if it's worth making a type of `FinalizedOrderStates` and `PendingOrderStates` to maintain consistency throughout the codebase (it might be a lil difficult cuz we use string literals, but I think even also having `[Finalized/Pending]OrderStateStrings` could be a reasonable in between state

moo-onthelawn (approved):
ty for the tests; some considers but nothing blocking

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/OrderProcessor.kt
```diff
@@ -303,11 +319,21 @@ internal class OrderProcessor(parser: ParserProtocol) : BaseProcessor(parser) {
         return Pair(existing, false)
     }
 
+    private fun isStatusFinalized(status: Any?): Boolean {
```
ruixhuang, line 322: Nit...  param can be typed as `status: String?`

ruixhuang (approved):

PR #448, opened 2024-06-17
Fix toInt() crash

ruixhuang:
I can't find in the current code state where the "null" key can be added.  Maybe the root cause has been fixed since the last Android build.  Nevertheless adding logic to check for "null" key in the subaccounts map.


src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
```diff
@@ -1366,23 +1366,6 @@ open class TradingStateMachine(
         )
     }
 
-    private fun calculateAccount(subaccountNumbers: IList<Int>, period: CalculationPeriod) {
```
prashanDYDX, line 1369: unused?

aforaleka (approved):

jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
ruixhuang: Correct

PR #452, opened 2024-06-17
Automatically cancel trigger orders if position has been closed/flipped

aforaleka:
- add logic to cancel trigger orders when its associated position has been closed or flipped
- tested locally and works on both cross and isolated positions' orphaned trigger orders

An order will be canceled if it meets the following conditions:
- order is a conditional order, reduce-only
- order is untriggered or open status
- order's market has no position or position was flipped (i.e. different direction than the order's side)

followup:
- currently cancel reason will say user canceled it, not ideal but there'll be extra logic needed to override indexer's cancel reason and we want to ship this asap
- tests :)

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
@@ -813,7 +813,6 @@ internal class SubaccountSupervisor(
         analyticsPayload: IMap<String, Any>?,
         uiClickTimeMs: Double,
         fromSlTpDialog: Boolean = false,
-        isOrphanedTriggerOrder: Boolean = false,
     ): HumanReadableCancelOrderPayload {
         val clientId = payload.clientId
         val string = Json.encodeToString(payload)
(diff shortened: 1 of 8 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -778,7 +812,8 @@ internal class SubaccountSupervisor(
         payload: HumanReadableCancelOrderPayload,
         analyticsPayload: IMap<String, Any>?,
         uiClickTimeMs: Double,
-        isTriggerOrder: Boolean = false,
+        fromSlTpDialog: Boolean = false,
+        isOrphanedTriggerOrder: Boolean = false,
```
moo-onthelawn, line 816: hmm do we use `isOrphanedTriggerOrder` here?

jaredvu (approved):

moo-onthelawn (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
aforaleka: gr8888 catch

PR #458, opened 2024-06-19
Compliance: Poll geo, deprecate ComplianceAction.Onboard

rosepuppy:


src/commonTest/kotlin/exchange.dydx.abacus/app/manager/v2/V4ForegroundCycleTests.kt
```diff
@@ -693,12 +693,12 @@ class V4ForegroundCycleTests : NetworkTests() {
                     "https://api.examples.com/configs/exchanges.json",
                     "https://api.dydx.exchange/v4/geo",
                     "https://indexer.v4staging.dydx.exchange/v4/screen?address=cosmos1fq8q55896ljfjj7v3x0qd0z3sr78wmes940uhm",
+                    "https://indexer.v4staging.dydx.exchange/v4/compliance/screen/cosmos1fq8q55896ljfjj7v3x0qd0z3sr78wmes940uhm",
```
yogurtandjam, line 696: [q] is this based on a mocked timer? or is this actually cycling based on our polling? 👀 i don't imagine a single test will take > 10 seconds but it would be really nice to not have to worry about timers for tests

cc: @prashanDYDX

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -751,12 +749,16 @@ internal open class AccountSupervisor(
...
+    internal open fun triggerCompliance(action: ComplianceAction, callback: TransactionCallback?) {
         if (compliance.status != ComplianceStatus.UNKNOWN) {
             updateCompliance(DydxAddress(accountAddress), compliance.status, action)
-            callback(true, null, null)
+            if (callback != null) {
+                callback(true, null, null)
+            }
+        }
+        if (callback != null) {
+            callback(false, V4TransactionErrors.error(null, "No account address"), null)
```
yogurtandjam, line 760: [discussion] i feel like since we can't use named args for passed in callback functions, it might be nice in the future to create a `{functionName}Args` dataclass so the fn signatures remain explicit 🤔

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/StateManagerAdaptorV2.kt
```diff
@@ -707,5 +717,10 @@ internal class StateManagerAdaptorV2(
                 ),
             )
         }
+        triggerCompliance(ComplianceAction.CONNECT, null)
+    }
+
+    companion object {
+        private const val GEO_POLLING_DURATION_SECONDS = 10.0
```
yogurtandjam, line 724: [q] why a private val instead of a global const?

yogurtandjam (approved):
LGTM

non-blocking - is there a way we can test this? we should be able to manipulate timing via https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/ 

it'll probably take a bit of time to set up and IMO worth making another PR for, but may be useful considering we have a few polling methods in our codebase

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -751,12 +749,16 @@ internal open class AccountSupervisor(
...
-    internal open fun triggerCompliance(action: ComplianceAction, callback: TransactionCallback) {
+    internal open fun triggerCompliance(action: ComplianceAction, callback: TransactionCallback?) {
         if (compliance.status != ComplianceStatus.UNKNOWN) {
             updateCompliance(DydxAddress(accountAddress), compliance.status, action)
-            callback(true, null, null)
+            if (callback != null) {
+                callback(true, null, null)
+            }
+        }
+        if (callback != null) {
```
ruixhuang, line 759: Is this correct?  The callback gets called twice when `compliance.status != ComplianceStatus.UNKNOW`

rosepuppy: Oh good catch, fixing

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/StateManagerAdaptorV2.kt
rosepuppy: could probably make this a global const 👍

PR #460, opened 2024-06-19
feat: skip status [OTE-349] [1/n]

yogurtandjam:
This looks like a heckin chonker but don't be fooled ~1500 LoC are from tests and mocks. It's closer to 200 LoC

Followed up by https://github.com/dydxprotocol/v4-abacus/pull/461

- creates skipStatusProcessor
- follows spec according to: https://www.notion.so/dydx/Squid-Skip-Status-Migration-7799a0902f124e2f980d9694f2fa2a23#93e956f20f9d4567b00f96fa5a8746d9
- transforms skip status response to be as equivalent as possible to squid status payload

PR #461, opened 2024-06-19
feat: skip status [OTE-349] [2/n]

yogurtandjam:
Followup to https://github.com/dydxprotocol/v4-abacus/pull/460

Fully unlocks the status tracking flow by calling the required `track` API call first before attempting to call `status`

- creates skipTrackProcessor
- adds new piece of abacus state trackStatuses
- simple bool switch to track whether or not a tx has been successfully tracked or not, so we can begin hitting it with `status`


src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
```diff
@@ -1067,6 +1078,7 @@ open class TradingStateMachine(
         var configs = state?.configs
         var input = state?.input
         var transferStatuses = state?.transferStatuses?.toIMutableMap()
+        var trackStatuses = state?.trackStatuses?.toIMutableMap()
```
prashanDYDX, line 1081: why does this need to be an `IMutableMap`? (`IMutableMap` is only needed for types that are consumed directly by JavaScript, internally in abacus we can just use `Map` and `.toMap()`

src/commonTest/kotlin/exchange.dydx.abacus/tests/payloads/SkipTrackMock.kt
```diff
@@ -0,0 +1,14 @@
+package exchange.dydx.abacus.tests.payloads
+
+class SkipTrackMock {
```
prashanDYDX, line 3: this doesn't need to be a class, can just be an `object`. then you don't need to instantiate this in your test class. can just access like `SkipTrackMock`

src/commonMain/kotlin/exchange.dydx.abacus/output/PerpetualState.kt
```diff
@@ -33,6 +33,7 @@ data class PerpetualState(
     val input: Input?,
     val availableSubaccountNumbers: IList<Int>,
     val transferStatuses: IMap<String, TransferStatus>?,
+    val trackStatuses: IMap<String, Boolean>?,
```
prashanDYDX, line 36: does this need to be nullable? does empty work here?

src/commonTest/kotlin/exchange.dydx.abacus/tests/payloads/SkipTrackMock.kt
yogurtandjam: got it - though FYI i've been following existing patterns so we should probably refactor existing mocks

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
yogurtandjam: i was under the impression that this state is exposed to the js client, no? this is the `update` fn that constructs the `PerpetualState` object so all the maps are `IMap`s

src/commonMain/kotlin/exchange.dydx.abacus/output/PerpetualState.kt
yogurtandjam: hmm... maybe. right now the abacus pattern is to return null for invalid data but we can probably change the default value to an empty map. i'll attempt to do it and timebox it for ~30 mins

yogurtandjam: ah - everything is pulled from the typeless `data` map so everything **has** to be nullable until we get state schematized

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
yogurtandjam: or are we intentionally selective about what abacus state items get exposed to the JS client

src/commonMain/kotlin/exchange.dydx.abacus/output/PerpetualState.kt
```diff
@@ -119,4 +120,8 @@ data class PerpetualState(
     fun transferStatus(hash: String): TransferStatus? {
         return transferStatuses?.get(hash)
     }
+
+    fun trackStatus(hash: String): Boolean? {
+        return trackStatuses?.get(hash)
+    }
```
mike-dydx, line 126: is this just a convenience method for map access? Ideally we access maps uniformly throughout the app IMO

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+Squid.kt
```diff
@@ -87,3 +87,16 @@ internal fun TradingStateMachine.squidStatus(
...
+
+internal fun TradingStateMachine.routerTrack(
+    payload: String,
+): StateChanges? {
+    val json = try {
+        Json.parseToJsonElement(payload).jsonObject.toMap()
+    } catch (exception: SerializationException) {
+        Logger.e { "Failed to deserialize skipTrack: $payload \nException: $exception" }
+        return StateChanges(iEmptyList())
+    }
```
mike-dydx, line 99: maybe can use ParserProtocol's

`decodeJsonObject` to reduce this bit

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -1158,6 +1163,48 @@ internal class OnboardingSupervisor(
...
+            return
+        }
+        val params: IMap<String, String> = iMapOf(
+            "tx_hash" to hash,
+            "chain_id" to fromChainId,
+        ).filterNotNull()
+        val url = helper.configs.skipV2Status()
+        helper.get(url, params) { _, response, httpCode, _ ->
+            if (response != null) {
+                update(stateMachine.squidStatus(response, hash), oldState)
```
mike-dydx, line 1184: squidstatus?

yogurtandjam: man, i think this is a rebase issue. this is an old PR. thanks

src/commonTest/kotlin/exchange.dydx.abacus/tests/payloads/SkipTrackMock.kt
```diff
@@ -0,0 +1,14 @@
+package exchange.dydx.abacus.tests.payloads
+
+object SkipTrackMock {
+    val error = """{
+  "code": 5,
+  "message": "chain id not supported",
+  "details": []
```
ruixhuang, line 7: Do we want to surface the error to the user? @mike-dydx

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+Squid.kt
yogurtandjam: decodeJsonObject performs an extra conversion to `IMap` which is slower than the native kotlin map. if you want i can make a `decodeJsonObjectToNativeMap` or something

src/commonTest/kotlin/exchange.dydx.abacus/tests/payloads/SkipTrackMock.kt
yogurtandjam: FWIW i'm just following the same error pattern we have for other router requests for now.

In the future for error messages we'll be making an abstraction layer to make these more end-user friendly

ruixhuang (approved):

PR #463, opened 2024-06-20
fix: simulate withdrawal with proper denom

yogurtandjam:
grab denom from the simulation payload and use the correct denom.

NOTE: our error messages don't show up properly in the UI if we lack gas. just says "Unknown error" 👀

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
@@ -525,7 +525,7 @@ internal class OnboardingSupervisor(
                     helper.parser.asDouble(state.input.transfer.size?.usdcSize)
                         ?: Numeric.double.ZERO
                 if (usdcSize > Numeric.double.ZERO) {
-                    simulateWithdrawal(decimals, subaccountNumber) { gasFee ->
+                    simulateWithdrawal(subaccountNumber) { gasFee ->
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -621,8 +620,13 @@ internal class OnboardingSupervisor(
...
             if (result != null) {
+                val usdcDecimals = helper.environment.tokens["usdc"]?.decimals ?: 6
+                val nativeDecimals = helper.environment.tokens["chain"]?.decimals ?: 18
+                val nativeDenom = helper.environment.tokens["chain"]?.denom
                 val amountMap =
                     helper.parser.asMap(helper.parser.asList(result["amount"])?.firstOrNull())
+                val denom = helper.parser.asString(amountMap?.get("denom"))
+                val decimals = if (denom === nativeDenom) nativeDecimals else usdcDecimals
                 val amount = helper.parser.asDecimal(amountMap?.get("amount"))
                 val usdcAmount = amount?.div(Numeric.decimal.TEN.pow(decimals))
```
jaredvu, line 631: `usdcAmount` here would be populated incorrectly even with the correct decimals.

In a later comparison, f the user is using `adydx` as their denom `usdcAmount`, it should not be subtracted from the `amount` input for validation

yogurtandjam:
closing b/c forward fixing this is a significantly higher lift than a quick hotfix

PR #465, opened 2024-06-20
Apply the same logic to calculate marginMode and targetLeverage for new and existing trade

ruixhuang:
Otherwise, when the first market to trade is isolated (i.e., initiateTrade() gets called), the targetLeverage of the position is not applied.


src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+TradeInput.kt
```diff
@@ -89,7 +89,13 @@ internal fun TradingStateMachine.tradeInMarket(
             // If we changed market, we should also reset the price and size
             modified.safeSet("size", null)
             modified.safeSet("price", null)
-
+            modified
```
mike-dydx, line 92: can remove

mike-dydx (approved):

PR #467, opened 2024-06-21
update margin transfer calculation with absolute net size (flipped isolated position)

aforaleka:
in calculating how much margin to transfer, should take absolute net size change instead of using the full order amount, so when the position is flipped, it doesn't transfer too much margin

image

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
@@ -6,6 +6,7 @@ import exchange.dydx.abacus.utils.Logger
 import exchange.dydx.abacus.utils.MAX_LEVERAGE_BUFFER_PERCENT
 import exchange.dydx.abacus.utils.MAX_SUBACCOUNT_NUMBER
 import exchange.dydx.abacus.utils.NUM_PARENT_SUBACCOUNTS
+import exchange.dydx.abacus.utils.Numeric
 import kollections.iListOf
 import kotlin.math.max
 import kotlin.math.min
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
```diff
@@ -280,14 +295,17 @@ internal object MarginCalculator {
...
+
         return if (adjustedTargetLeverage == 0.0 || adjustedTargetLeverage == null) {
             null
         } else {
             getTransferAmountFromTargetLeverage(
                 price,
                 oraclePrice,
                 side,
-                size,
+                positionSizeDifference,
```
jaredvu, line 308: :tada:

src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
```diff
@@ -202,15 +203,13 @@ internal object MarginCalculator {
         subaccount: Map<String, Any>?,
         tradeInput: Map<String, Any>?,
     ): Boolean {
-        val marketId = parser.asString(tradeInput?.get("marketId")) ?: return true
-        val position = parser.asNativeMap(parser.value(subaccount, "openPositions.$marketId"))
-        val currentSize = parser.asDouble(parser.value(position, "size.current")) ?: 0.0
-        val postOrderSize = parser.asDouble(parser.value(position, "size.postOrder")) ?: 0.0
-        return postOrderSize.abs() > currentSize.abs()
+        return getPositionSizeDifference(parser, subaccount, tradeInput)?.let {
+            it.compareTo(Numeric.double.ZERO) == 1
```
jaredvu, line 207: From the kotlin [docs](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-comparable/compare-to.html), i can't tell if the comparison to 1 will work. would it be safer to `> 0` in this case?

jaredvu: if we know it will be 1 or -1 then I defer to you 😄

jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
```diff
@@ -204,7 +204,7 @@ internal object MarginCalculator {
         tradeInput: Map<String, Any>?,
     ): Boolean {
         return getPositionSizeDifference(parser, subaccount, tradeInput)?.let {
-            it.compareTo(Numeric.double.ZERO) == 1
+            it.compareTo(Numeric.double.ZERO) > 0 // whether it's is positive
```
jaredvu, line 207: nit: can remove `'s`

PR #468, opened 2024-06-22
Add marginValue to SubaccountPosition

ruixhuang:
We use this to display the margin for each position.  Currently we do this calculation at client, so let's move it to Abacus.


jaredvu (approved):
can we also add the new fields to the SubaccountPosition markdown file?

PR #469, opened 2024-06-24
Fix negative input to margin input amount

ruixhuang:
Mobile app sends the input field as-is, so there can be negative numbers.  For margin, we will just ignore the sign.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -51,7 +51,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.8.0"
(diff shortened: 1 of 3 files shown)


yogurtandjam (approved):
Is there an easy way to unit test this? Would be nice to have a simple test to make sure the margin sign gets ignored

jaredvu (approved):

ruixhuang:
> Is there an easy way to unit test this? Would be nice to have a simple test to make sure the margin sign gets ignored

👍  Added

PR #471, opened 2024-06-25
feat: expose get chain by id method

yogurtandjam:
Exposes get chain by id helper method. This allows us to get the `chainName` that the frontend wants to display. This is also probably just useful in general. 

This poses the question - should we continue being 'react'y and storing all react related data into the state? Or do we want to expose data that react can't update directly via abacus.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
@@ -145,7 +145,7 @@ internal class SkipProcessor(
         internalState.tokenResources = tokenResources(selectedChainId)
     }
 
-    private fun getChainById(chainId: String): Map<String, Any>? {
+    override fun getChainById(chainId: String): Map<String, Any>? {
         return parser.asNativeMap(this.chains?.find { parser.asString(parser.asNativeMap(it)?.get("chain_id")) == chainId })
     }
 
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
@@ -145,6 +145,8 @@ internal class SkipProcessor(
         internalState.tokenResources = tokenResources(selectedChainId)
     }
 
+//    Update this when we move chains to state
+//    https://linear.app/dydx/issue/OTE-470/migrate-chains-and-tokens-to-abacus-state
     override fun getChainById(chainId: String): Map<String, Any>? {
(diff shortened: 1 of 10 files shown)


v4_abacus.podspec
```diff
@@ -11,7 +11,7 @@ Pod::Spec.new do |spec|
                 
                 
                 
-    if false
+    if !Dir.exist?('build/cocoapods/framework/Abacus.framework') || Dir.empty?('build/cocoapods/framework/Abacus.framework')
```
ruixhuang, line 14: @yogurtandjam Can you revert this line?  Shouldn't need to update this.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/utils/Payloads.kt
```diff
@@ -130,3 +130,13 @@ data class HumanReadableTransferPayload(
     val amount: String,
     val recipient: String,
 )
+
+@JsExport
+@Serializable
+data class HumanReadableChainPayload(
```
ruixhuang, line 136: `HumanReadablePayload` typically refers to the payload to be submitted with on-chain calls (i.e., calls to v4-clients functions).  Since this is just a normal struct,  can we remove rename this to something like `TransferChainInfo`?

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
```diff
@@ -145,7 +145,9 @@ internal class SkipProcessor(
         internalState.tokenResources = tokenResources(selectedChainId)
     }
 
-    private fun getChainById(chainId: String): Map<String, Any>? {
+//    Update this when we move chains to state
+//    https://linear.app/dydx/issue/OTE-470/migrate-chains-and-tokens-to-abacus-state
```
ruixhuang, line 149: It should be easy just to add the `chains` into `PerpetualState`.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/utils/Payloads.kt
yogurtandjam: sounds good tom e!

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
yogurtandjam: yeah it shouldn't be too difficult - just thought it made sense logically to do it when i migrate the state to use the new `InternalState` struct

yogurtandjam: synced on slack: fill update this after landing PR to main

ruixhuang (approved):

PR #475, opened 2024-06-25
Isolated markets: reduce position delta

aforaleka:
when inputting a trade (or close position) that reduces an isolated position, and there is no other open orders, postorder cross free collateral should increase, and position margin for the existing isolated position should decrease

main changes:
- introduces the concept of how much collateral can be transferred out (which we'll later incorporate to transfer out collateral on subaccount/ positions change); this will allow cross free collateral to reflect post transfer amount
- fixes usdcSize (which can be the transfer amount) being overwritten in `transformDelta` by passing in extra param `hasTransfer` that skips that

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -51,7 +51,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.8.6"
+version = "1.8.7"
 
 repositories {
     google()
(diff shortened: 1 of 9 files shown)


ruixhuang:
The tests are so complicated 😓

ruixhuang (approved):

PR #477, opened 2024-06-26
fix: proper jsonification of errors + enabling dynamic skip url

yogurtandjam:
errors were showing up as `unknown error` because kotlin distinguishes between JSON array and regular array.

skipHost was hard coded. now it accepts a dynamic value from the env.json so DOS can use their proxied server to inject their API key


ruixhuang (approved):

moo-onthelawn (approved):
~~lgtm, tho it looks like maybe we bumped version twice?~~ oh jk i seee u just didn't pull in the most upgraded gradle

PR #478, opened 2024-06-26
isolated [2/n]: calculate transfer payload for isolated trade on the fly

aforaleka:
+ refactoring
+ calculate transfer payload for isolated trade on placing order
+ close position should get a more efficient way to find the child subaccount number (can assume it's the first subaccount with that open market position)

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
@@ -218,6 +218,17 @@ internal object MarginCalculator {
         error("No available subaccount number")
     }
 
+    fun getChildSubaccountNumberForIsolatedMarginClosePosition(
+        parser: ParserProtocol,
+        account: Map<String, Any>?,
+        subaccountNumber: Int,
+        tradeInput: Map<String, Any>?
+    ): Int {
+        val marketId = parser.asString(tradeInput?.get("marketId")) ?: return subaccountNumber
+        val position = findExistingPosition(parser, account, marketId, subaccountNumber)
+        return parser.asInt(position?.get("subaccountNumber")) ?: subaccountNumber
+    }
+
     /**
      * @description Calculate and validate the amount of collateral to transfer for an isolated margin trade.
      */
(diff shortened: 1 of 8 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
```diff
@@ -237,18 +251,29 @@ internal object MarginCalculator {
...
-    private fun getIsIncreasingPositionSize(
-        parser: ParserProtocol,
-        subaccount: Map<String, Any>?,
-        tradeInput: Map<String, Any>?,
-    ): Boolean {
-        return getPositionSizeDifference(parser, subaccount, tradeInput)?.let { it > 0 } ?: true
+    /**
+     * @description Calculate the amount of collateral to transfer into child subaccount for an isolated margin trade.
+     */
+    fun getIsolatedMarginTransferInAmountForTradeTyped(
```
jaredvu, line 257: nit: `Typed` -> `Type`

src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
```diff
@@ -372,7 +411,52 @@ internal object MarginCalculator {
...
+            side,
+            oraclePrice,
+            price,
+            initialMarginFraction,
+            effectiveImf,
+            positionSizeDifference,
+        )
+    }
+
+    private fun calculateIsolatedMarginTransferAmountTyped(
```
jaredvu, line 427: nit: `Typed` -> `Type`

src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
aforaleka: mmmm i mean that the inputs are typed, that's why i added `Typed` at the end -- is there a better naming that could suggest that?

aforaleka: or i can just have the same name which is kinda ok since the caller can infer based on inputs

jaredvu (approved):

PR #481, opened 2024-06-27
isolated: fix transfer out condition

aforaleka:
reduce position -> fully closed position

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
@@ -6,7 +6,6 @@ import exchange.dydx.abacus.utils.Logger
 import exchange.dydx.abacus.utils.MAX_LEVERAGE_BUFFER_PERCENT
 import exchange.dydx.abacus.utils.MAX_SUBACCOUNT_NUMBER
 import exchange.dydx.abacus.utils.NUM_PARENT_SUBACCOUNTS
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
```diff
@@ -263,21 +260,34 @@ internal object MarginCalculator {
...
-     * @description Determine if collateral should be transferred out for an isolated margin trade
+     * @description Determine if collateral should be transferred out from child subaccount for an isolated margin trade
      */
     fun getShouldTransferOutCollateral(
         parser: ParserProtocol,
         subaccount: Map<String, Any>?,
         tradeInput: Map<String, Any>?,
     ): Boolean {
-        val isDecreasingPositionSize = !getIsIncreasingPositionSize(parser, subaccount, tradeInput)
+        val isPositionFullyClosed = getIsPositionFullyClosed(parser, subaccount, tradeInput)
```
jaredvu, line 270: wouldn't we still want to check is decreasingPositionSize as well? This was to not transfer if they flipped sides but their net position was decreasing.

jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
aforaleka: resolved offline -- this is a new function that was added for the diff state and fixed logic is the right one!

PR #482, opened 2024-06-27
isolated [1/n]: update position size difference calculation / fix slow receipt

aforaleka:
calculate how much to transfer with position size difference inferred from trade input instead of position, since delta has not been applied to position
this should fix receipt being slow / weird state where it was using an outdated position size

alternative route of applying transfer after position delta is duplicating even more code so didn't go down that path

https://github.com/dydxprotocol/v4-abacus/assets/9400120/3d31023b-1ac3-4e48-81e9-6f0393f7ae83

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
@@ -1,12 +1,15 @@
 package exchange.dydx.abacus.calculator
 
 import abs
+import exchange.dydx.abacus.output.PerpetualMarket
+import exchange.dydx.abacus.output.Subaccount
+import exchange.dydx.abacus.output.input.MarginMode
+import exchange.dydx.abacus.output.input.TradeInput
 import exchange.dydx.abacus.protocols.ParserProtocol
 import exchange.dydx.abacus.utils.Logger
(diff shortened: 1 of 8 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginCalculator.kt
```diff
@@ -301,11 +302,37 @@ internal object MarginCalculator {
...
+        trade: Map<String, Any>,
+        currentPositionSize: Double,
+    ): Double {
+        val marketId = parser.asString(trade["marketId"])
+        val side = parser.asString(trade["side"])
+        val tradeSize = if (marketId != null && side != null) {
+            parser.asNativeMap(trade["summary"])?.let { summary ->
+                if (parser.asBool(summary["filled"]) == true) {
+                    val multiplier = if (side == "BUY") Numeric.double.NEGATIVE else Numeric.double.POSITIVE
+                    (parser.asDouble(summary["size"]) ?: Numeric.double.ZERO) * multiplier * Numeric.double.NEGATIVE
```
jaredvu, line 325: qq: Why the `multiplier * NEGATIVE`? Why not just switch `multiplier` and remove this `NEGATIVE`?

aforaleka: copied from delta code but u right!! we don't need to do that here, will fix

jaredvu (approved):

PR #486, opened 2024-06-28
isolated: attempt reclaim funds on subaccount change instead of polling

aforaleka:
instead of polling reclaim funds from child subaccounts:
- subscribe to subaccount change
- when that happens, attempt to reclaim funds if there are 1. no open positions, 2. no open orders, 3. no in progress placing orders
- 3 should prevent the case where trader is placing a new isolated order to open a new position but funds get swept back

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
@@ -1492,7 +1492,7 @@ internal class SubaccountSupervisor(
     /**
      * @description Loop through all subaccounts to find childSubaccounts that have funds but no open positions or orders. Initiate a transfer to parentSubaccount.
      */
-    private fun reclaimUnutilizedFundsFromChildSubaccounts() {
+    internal fun reclaimUnutilizedFundsFromChildSubaccounts() {
         val subaccounts = stateMachine.state?.account?.subaccounts ?: return
 
         val subaccountQuoteBalanceMap = subaccounts.mapValues { subaccount ->
@@ -1517,7 +1517,8 @@ internal class SubaccountSupervisor(
             // this is to check the case where we're still waiting for subaccount transfer to complete
             // before the isolated market order can be placed
             val isPlacingOrderForSubaccount = placeOrderRecords.any {
-                it.destinationSubaccountNumber == subaccount.value.subaccountNumber
+                it.destinationSubaccountNumber == subaccount.value.subaccountNumber &&
+                        it.lastOrderStatus == null // i.e. not indexed, we let `hasIndexedOpenOrder` be source of truth once order is indexed
             }
 
             // Only return a quoteBalance if the subaccount has no open positions or orders
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
@@ -1518,7 +1518,7 @@ internal class SubaccountSupervisor(
             // before the isolated market order can be placed
             val isPlacingOrderForSubaccount = placeOrderRecords.any {
                 it.destinationSubaccountNumber == subaccount.value.subaccountNumber &&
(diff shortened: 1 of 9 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -1487,13 +1483,16 @@ internal class SubaccountSupervisor(
         if (changes.changes.contains(Changes.subaccount)) {
             parseOrdersToMatchPlaceOrdersAndCancelOrders()
             cancelTriggerOrdersWithClosedOrFlippedPositions()
+            reclaimUnutilizedFundsFromChildSubaccounts()
         }
     }
 
+    private var reclaimingChildSubaccountNumbers = mutableSetOf<Int>()
```
prashanDYDX, line 1490: `val`

prashanDYDX: also can we move this to the top of the file?

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -1487,13 +1483,16 @@ internal class SubaccountSupervisor(
...
+
     /**
      * @description Loop through all subaccounts to find childSubaccounts that have funds but no open positions or orders. Initiate a transfer to parentSubaccount.
      */
-    private fun reclaimUnutilizedFundsFromChildSubaccounts() {
+    internal fun reclaimUnutilizedFundsFromChildSubaccounts() {
```
prashanDYDX, line 1495: is this `internal` just for testing?

prashanDYDX: it looks like this could be private actually right?

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -1544,46 +1550,28 @@ internal class SubaccountSupervisor(
...
-    }
-
-    private fun recursivelyReclaimChildSubaccountFunds(transferPayloadStrings: MutableList<String>) {
-        if (transferPayloadStrings.isNotEmpty()) {
-            val transferPayloadString = transferPayloadStrings.removeAt(0)
-            helper.transaction(TransactionType.SubaccountTransfer, transferPayloadString) { response ->
+        transferPayloads.forEach { transferPayload ->
+            val transferPayloadString = Json.encodeToString(transferPayload)
+            val transactionCallback = { response: String? ->
+                this.reclaimingChildSubaccountNumbers.remove(transferPayload.subaccountNumber)
```
prashanDYDX, line 1562: do we want to remove from set even if there was an error?

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
aforaleka: ah yea added a test for it but realized can just let the subaccount change trigger -- will change it back to private

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
aforaleka: yea i think so, will rather let it retry the transfer?

jaredvu (approved):

prashanDYDX (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
prashanDYDX: chatted offline, not retrying makes sense here - unclear if a retry would help the problem. let's add some error logging (maybe amplitude?) so we can track this

PR #492, opened 2024-07-01
Codegen model from indexer swagger

ruixhuang:
Run ./swagger_codegen,sh to generate the model classes from the swagger.json of the indexer repo.  There are still some issues within the swagger.json that we should work with the Indexer team to resolve.  But for now the models generated give us a starting point to work with.

I will follow up a PR to start migrating the processor logic using the codgen result.


aforaleka:
im excited for dis

prashanDYDX:
is there a way to configure the generated models to have all nullable types? since breaking api changes seem not uncommon, probably want to make sure deserialization is defensive.

ruixhuang:
> is there a way to configure the generated models to have all nullable types? since breaking api changes seem not uncommon, probably want to make sure deserialization is defensive.

Good call.  There is no config option for this unfortunately.  By default, all fields are optional unless they are marked by the swagger.json as "required".  Typically, front-end facing API are backward compatible, and breaking change should be versioned (e.g., either with a different endpoint or with the client asking for a different version).  In the worst case, we can catch serialization error and try to surface the force-update screen to the user if that happens.   Open to other ideas.

ruixhuang:
Updated with optional fields

src/commonMain/kotlin/indexer/codegen/IndexerAPIOrderStatus.kt
```diff
@@ -0,0 +1,20 @@
...
+ */
+package indexer.codegen
+
+import kotlinx.serialization.Serializable
+
+/**
+ *
+ */
+@Serializable
+class IndexerAPIOrderStatus()
```
prashanDYDX, line 20: what are these types with no fields?

prashanDYDX (approved):

swagger_codegen.sh
```diff
@@ -0,0 +1,40 @@
...
+# Function to clean up the temporary directory
+cleanup() {
+    echo "Cleaning up..."
+    rm -rf "$TMP_DIR"
+}
+
+# Trap to clean up in case of script exit or interruption
+trap cleanup EXIT
+
+curl -o $TMP_DIR/swagger.json https://raw.githubusercontent.com/dydxprotocol/v4-chain/ce926099af057e38d7d79453d01c5ebe0cdc3634/indexer/services/comlink/public/swagger.json
```
prashanDYDX, line 17: probably want to update this url to use `main`:
https://raw.githubusercontent.com/dydxprotocol/v4-chain/main/indexer/services/comlink/public/swagger.json

prashanDYDX:
fyi looks like the framework indexer uses to generate swagger.json migrated from Swagger to OpenAPI back in 2020. I _think_ this should still be compatible but something to be aware of: https://github.com/lukeautry/tsoa/pull/664

PR #497, opened 2024-07-02
Fix compliance screen discrepancy

rosepuppy:


yogurtandjam (approved):
stamped for quick bugfix, but if it's not too hard (i know it's impossible to unit test some functions), something with nuanced logic probably merits a unit test

PR #502, opened 2024-07-03
Refactor v4_markets subscribed utilizing data classes

jaredvu:
Add a different format to the internal data that we are working with.
- It just consumed Indexer v4_markets subscribed message outright and saves into a data-class
- Next steps would be to add the derived data, fix tests, and support channel_batch_data updates


src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/MarketsSummaryProcessor.kt
```diff
@@ -24,6 +28,70 @@ internal class MarketsSummaryProcessor(parser: ParserProtocol, calculateSparklin
         return modify(existing, markets)
     }
 
+    internal fun testSubscribed(content: Map<String, Any>): InternalStatePerpetualMarkets {
+        val payload = parser.asNativeMap(content["markets"])
```
ruixhuang, line 32: With Indexer codegen already landed, we can use static types for input payload with something like this now:

`        val marketResponse = Json.decodeFromString<IndexerPerpetualMarketResponse?>(content.toJson())`

PR #503, opened 2024-07-06
 Static typing for subaccount fills

ruixhuang:
Constructing subaccount fills with static typed data and codegen models from Indexer.  This can serve as a reference for future migration.

- Feature-flagged via `staticTyping` from Abacus config and retaining all existing behavior with the data map.   
- New processor functions have names starting with "process".
- Previous data map-based functions are renamed with trailing "Deprecated".  
- Tested with Android.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/WalletProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/WalletProcessor.kt
@@ -3,19 +3,37 @@ package exchange.dydx.abacus.processor.wallet
 import exchange.dydx.abacus.processor.base.BaseProcessor
 import exchange.dydx.abacus.processor.wallet.account.V4AccountProcessor
 import exchange.dydx.abacus.processor.wallet.user.UserProcessor
+import exchange.dydx.abacus.protocols.LocalizerProtocol
 import exchange.dydx.abacus.protocols.ParserProtocol
 import exchange.dydx.abacus.responses.SocketInfo
+import exchange.dydx.abacus.state.internalstate.InternalWalletState
 import exchange.dydx.abacus.state.manager.BlockAndTime
 import exchange.dydx.abacus.utils.mutable
 import exchange.dydx.abacus.utils.safeSet
+import indexer.codegen.IndexerFillResponseObject
 
 internal class WalletProcessor(
     parser: ParserProtocol,
+    localizer: LocalizerProtocol?,
 ) : BaseProcessor(parser) {
-    private val v4accountProcessor = V4AccountProcessor(parser = parser)
+    private val v4accountProcessor = V4AccountProcessor(parser = parser, localizer = localizer)
     private val userProcessor = UserProcessor(parser = parser)
 
-    internal fun subscribed(
+    internal fun processSubscribed(
+        existing: InternalWalletState,
+        content: Map<String, Any>,
+        height: BlockAndTime?,
+    ): InternalWalletState {
+        val account = v4accountProcessor.processSubscribed(
+            existing = existing.account,
+            content = content,
+            height = height,
+        )
+        existing.account = account
+        return existing
+    }
+
+    internal fun subscribedDeprecated(
         existing: Map<String, Any>?,
         content: Map<String, Any>,
         height: BlockAndTime?,
@@ -26,13 +44,29 @@ internal class WalletProcessor(
             parser.asNativeMap(content),
         ) { existing, payload ->
             parser.asNativeMap(payload)?.let {
-                v4accountProcessor.subscribed(parser.asNativeMap(existing), it, height)
+                v4accountProcessor.subscribedDeprecated(parser.asNativeMap(existing), it, height)
             }
         }
     }
 
+    internal fun processChannelData(
+        existing: InternalWalletState,
+        content: Map<String, Any>,
+        info: SocketInfo,
+        height: BlockAndTime?,
+    ): InternalWalletState {
+        val account = v4accountProcessor.processChannelData(
+            existing = existing.account,
+            content = content,
+            info = info,
+            height = height,
+        )
+        existing.account = account
+        return existing
+    }
+
     @Suppress("FunctionName")
-    internal fun channel_data(
+    internal fun channel_dataDeprecated(
         existing: Map<String, Any>?,
         content: Map<String, Any>,
         info: SocketInfo,
@@ -204,13 +238,29 @@ internal class WalletProcessor(
         }
     }
 
-    internal fun receivedFills(
+    internal fun processFills(
+        existing: InternalWalletState,
+        payload: List<IndexerFillResponseObject>?,
+        subaccountNumber: Int,
+    ): InternalWalletState {
+        val newAccount = v4accountProcessor.processFills(
+            existing = existing.account,
+            payload = payload,
+            subaccountNumber = subaccountNumber,
+        )
+        if (newAccount != existing.account) {
+            existing.account = newAccount
+        }
+        return existing
+    }
+
+    internal fun receivedFillsDeprecated(
         existing: Map<String, Any>?,
         payload: Map<String, Any>,
         subaccountNumber: Int,
     ): Map<String, Any>? {
         return receivedObject(existing, "account", payload) { existing, payload ->
-            v4accountProcessor.receivedFills(
+            v4accountProcessor.receivedFillsDeprecated(
                 parser.asNativeMap(existing),
                 parser.asNativeMap(payload),
                 subaccountNumber,
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/FillProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/FillProcessor.kt
@@ -1,13 +1,24 @@
 package exchange.dydx.abacus.processor.wallet.account
 
+import exchange.dydx.abacus.output.FillLiquidity
+import exchange.dydx.abacus.output.SubaccountFill
+import exchange.dydx.abacus.output.SubaccountFillResources
 import exchange.dydx.abacus.output.input.MarginMode
+import exchange.dydx.abacus.output.input.OrderSide
+import exchange.dydx.abacus.output.input.OrderType
 import exchange.dydx.abacus.processor.base.BaseProcessor
 import exchange.dydx.abacus.processor.utils.OrderTypeProcessor
+import exchange.dydx.abacus.protocols.LocalizerProtocol
 import exchange.dydx.abacus.protocols.ParserProtocol
+import exchange.dydx.abacus.utils.Logger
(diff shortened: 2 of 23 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/WalletProcessor.kt
```diff
@@ -3,19 +3,37 @@ package exchange.dydx.abacus.processor.wallet
...
+        val account = v4accountProcessor.processSubscribed(
+            existing = existing.account,
+            content = content,
+            height = height,
+        )
+        existing.account = account
+        return existing
+    }
+
+    internal fun subscribedDeprecated(
```
prashanDYDX, line 36: kotlin-tip: there's an `@Deprecated` annotation we can use here too :)

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
```diff
@@ -49,24 +58,54 @@ internal open class SubaccountProcessor(parser: ParserProtocol) : BaseProcessor(
...
     }
 
-    internal open fun socket(
+    internal open fun processSocket(
+        existing: InternalSubaccountState,
+        content: Map<String, Any>,
+        subscribed: Boolean,
+        height: BlockAndTime?,
+    ): InternalSubaccountState {
+        val fillResponse = Json.decodeFromString<IndexerFillResponse?>(content.toJson())
```
prashanDYDX, line 100: need try catch here

prashanDYDX: Screenshot 2024-07-09 at 6 47 37 PM

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
```diff
@@ -238,15 +277,31 @@ internal open class SubaccountProcessor(parser: ParserProtocol) : BaseProcessor(
...
+        subaccount: InternalSubaccountState,
+        payload: List<IndexerFillResponseObject>?,
+        reset: Boolean,
+    ): InternalSubaccountState {
+        val newFills = fillsProcessor.process(
+            existing = if (reset) null else subaccount.fills,
+            payload = payload ?: emptyList(),
+            subaccountNumber = subaccount.subaccountNumber,
+        )
+        if (subaccount.fills != newFills) {
```
prashanDYDX, line 290: do we need this check here? why not just always set?

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/FillProcessor.kt
```diff
@@ -61,7 +79,59 @@ internal class FillProcessor(parser: ParserProtocol) : BaseProcessor(parser) {
         "SELL" to "Sell",
     )
 
-    internal fun received(
+    override fun process(
+        payload: IndexerFillResponseObject,
+        subaccountNumber: Int,
+    ): SubaccountFill? {
+        fun doProcess(): SubaccountFill? {
+            val fillSubaccountNumber = parser.asInt(payload.subaccountNumber) ?: subaccountNumber
```
prashanDYDX, line 87: don't think we need Parser.toInt here

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/FillsProcessor.kt
```diff
@@ -1,18 +1,42 @@
...
+    private val fillProcessor: FillProcessorProtocol = FillProcessor(parser = parser, localizer = localizer),
+) : BaseProcessor(parser) {
 
-    fun received(existing: List<Any>?, payload: List<Any>, subaccountNumber: Int): List<Any>? {
+    fun process(
+        existing: List<SubaccountFill>?,
+        payload: List<IndexerFillResponseObject>,
+        subaccountNumber: Int
+    ): List<SubaccountFill> {
+        val new = payload.mapNotNull { eachPayload ->
```
prashanDYDX, line 21: `mapNotNull` not necessary here right? isn't `payload` a List of non-nullable IndexerFillResponseObject?

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
```diff
@@ -238,15 +277,31 @@ internal open class SubaccountProcessor(parser: ParserProtocol) : BaseProcessor(
         return Pair(existing, false)
     }
 
-    private fun receivedFills(
+    private fun processFills(
+        subaccount: InternalSubaccountState,
+        payload: List<IndexerFillResponseObject>?,
```
prashanDYDX, line 282: can we make this non-nullable and handle nullability at call site?

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+Wallet.kt
```diff
@@ -162,9 +171,14 @@ internal fun TradingStateMachine.receivedFills(
     payload: Map<String, Any>,
     subaccountNumber: Int,
 ): StateChanges {
-    val size = parser.asList(payload["fills"])?.size ?: 0
+    val fills = parser.asList(payload["fills"])
+    val size = fills?.size ?: 0
     return if (size > 0) {
-        wallet = walletProcessor.receivedFills(wallet, payload, subaccountNumber)
+        wallet = walletProcessor.receivedFillsDeprecated(wallet, payload, subaccountNumber)
```
prashanDYDX, line 177: this can be in an `else` block right?

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
```diff
@@ -79,13 +79,13 @@ import kotlin.time.Duration.Companion.days
...
 @Suppress("UNCHECKED_CAST")
 @JsExport
-// @Serializable
 open class TradingStateMachine(
     private val environment: V4Environment?,
     private val localizer: LocalizerProtocol?,
     private val formatter: Formatter?,
     private val maxSubaccountNumber: Int,
     private val useParentSubaccount: Boolean,
+    val staticTyping: Boolean = false,
```
prashanDYDX, line 88: can be private i think?

v4_abacus.podspec
```diff
@@ -11,7 +11,7 @@ Pod::Spec.new do |spec|
                 
                 
                 
-    if false
+    if !Dir.exist?('build/cocoapods/framework/Abacus.framework') || Dir.empty?('build/cocoapods/framework/Abacus.framework')
```
prashanDYDX, line 14: i think you might need to rebase, mike had a fix for this i think

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+Wallet.kt
```diff
@@ -162,9 +171,14 @@ internal fun TradingStateMachine.receivedFills(
...
+        if (staticTyping) {
+            val payload = Json.decodeFromString<IndexerFillResponse?>(payload.toJson())
```
prashanDYDX, line 179: try catch

prashanDYDX: Jeremy had a good suggestion before that it might be useful to create a helper function around `Json.decodeFromString` with a try-catch. Then we can forbid usages of it directly via Detekt.

src/commonTest/kotlin/exchange.dydx.abacus/tests/mock/processor/wallet/account/FillProcessorMock.kt
```diff
@@ -0,0 +1,18 @@
+package exchange.dydx.abacus.tests.mock.processor.wallet.account
+
+import exchange.dydx.abacus.output.SubaccountFill
+import exchange.dydx.abacus.processor.wallet.account.FillProcessorProtocol
+import indexer.codegen.IndexerFillResponseObject
+
+class FillProcessorMock : FillProcessorProtocol {
```
prashanDYDX, line 7: super-nit: Personally I prefer the term `Fake` for implementations written by us. Colloquially, I understand `Mock` to refer to implementations generated by a mocking framework.

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+Wallet.kt
prashanDYDX: actually nevermind, this PR seems to only affect write path, not read path. so still need to write to the old map if i am understanding correctly.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: My follow-up PR already has this handled.

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+Wallet.kt
ruixhuang: Read should go through internal state as well.. but I don't think there is read on fills within Abacus

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
ruixhuang: There are extension functions reading it.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/FillsProcessor.kt
ruixhuang: It's necessary since item processor might return null (when required fields are missing).

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: Can update in followup PR

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: It's the same pattern we had before where we don't update the object unless there is a diff.  Downstream processing relies on this when constructing the change set for web FE.

prashanDYDX (approved):

PR #507, opened 2024-07-08
isolated: add additional guard against sweeping back funds when placing isolated orders

aforaleka:


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -765,7 +794,7 @@ internal class SubaccountSupervisor(
             submitTransaction(
                 TransactionType.SubaccountTransfer,
                 transferPayloadString,
-                null,
+                onSubmitIsolatedTransferTransaction,
```
jaredvu, line 797: 🎉

jaredvu (approved):

PR #508, opened 2024-07-10
add cache options to rest protocol

rosepuppy:
We need the ability to prevent [browser http caching](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache) for /geo. To do this we need to pass it to the RestProtocol:

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/protocols/PublicProtocols.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/protocols/PublicProtocols.kt
@@ -1,6 +1,7 @@
 package exchange.dydx.abacus.protocols
 
 import exchange.dydx.abacus.output.Asset
+import exchange.dydx.abacus.output.CandleOption
 import exchange.dydx.abacus.output.FeeTier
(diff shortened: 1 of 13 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/protocols/PublicProtocols.kt
```diff
@@ -80,32 +80,41 @@ interface RestProtocol {
...
+        options: RestOptions?,
         callback: RestCallback,
     )
 }
 
 typealias RestCallback = (response: String?, httpCode: Int, headersAsJsonString: String?) -> Unit
 
+@JsExport
+data class RestOptions(
+    val cache: String? = null
```
prashanDYDX, line 115: can this be an enum instead of a string?

yogurtandjam:
It looks like changes have to be made in many files and many functions which might be hard to upkeep, how do you feel (*if* we choose to continue doing this in the FE instead of BE) making a "no CacheFetch" function or something - that injects this option? 🤔

rosepuppy:
No longer needed as /geo updated Cache-control to no-store in the response.

PR #512, opened 2024-07-11
add ethRpc to env endpoints

mike-dydx:
as part of the squid -> skip migration, the skip chains API does not provide an rpc endpoint. This is because the skip team believes public rpc endpoints are not great. The Squid sdk API does provide an RPC endpoint though. There is a new rpc [configuration json ](https://github.com/dydxprotocol/v4-web/blob/main/public/configs/rpc.json)which this PR makes available


src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipChainResourceProcessor.kt
```diff
@@ -10,7 +11,7 @@ internal class SkipChainResourceProcessor(private val parser: ParserProtocol) {
     ): TransferInputChainResource {
         return TransferInputChainResource(
             chainName = parser.asString(payload["chain_name"]),
-            rpc = parser.asString(payload["rpc"]),
+            rpc = parser.asString(payload["chain_id"])?.let { RpcConfigs.chainIdToRpcMap[it]?.rpcUrl },
```
mike-dydx, line 14: @prashanDYDX wondering if there is a better place to put this than the processor. Doesn't feel right to look up a mapping in the processor since the mapping is part of the overall state. I feel like a processor should be declarative and not depend on context state :/

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -137,6 +141,17 @@ internal class OnboardingSupervisor(
         }
     }
 
+    private fun retrieveChainRpcEndpoints() {
+        val url = "${helper.deploymentUri}/configs/rpc.json"
+        helper.get(url) { _, response, _, _ ->
+            if (response != null) {
+                Json.decodeFromString<Map<String, RpcInfo>>(response).let {
+                    RpcConfigs.chainIdToRpcMap = it
```
ruixhuang, line 149: There can be a race-condition where the Skip endpoint returns before the config.json returns, in which the case the rcp will be null.  Might want to do the Skip resource fetch after this instead of in parallel.

yogurtandjam (commented):
Can we add a quick test for this? To make sure it's working as expected. I have a SkipProcessorTests file, you can create a SkipChainResourcesProcessorTests file

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/RpcInfo.kt
```diff
@@ -0,0 +1,15 @@
...
+
+@Serializable
+data class RpcInfo(
+    @SerialName("rpc") val rpcUrl: String,
+    val name: String,
+)
+
+
+object RpcConfigs {
+    var chainIdToRpcMap: Map<String, RpcInfo> = mapOf()
```
prashanDYDX, line 14: think we need a default map here in case the rpc request fails. which RPC endpoints do we actually need? the generated json in the web repo has so many chains (is this like every known EVM chain?).

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipChainResourceProcessor.kt
prashanDYDX: you could remove rpc as a property entirely here, replace it with chain id and do the look ups at usage sites. that way if the rpc configs update we don't have an outdated endpoints hanging around.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -137,6 +141,17 @@ internal class OnboardingSupervisor(
         }
     }
 
+    private fun retrieveChainRpcEndpoints() {
+        val url = "${helper.deploymentUri}/configs/rpc.json"
+        helper.get(url) { _, response, _, _ ->
+            if (response != null) {
+                Json.decodeFromString<Map<String, RpcInfo>>(response).let {
```
prashanDYDX, line 148: need a try-catch here: `decodeFromString` throws `SerializationException` and `IllegalArgumentException`

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
prashanDYDX: if we do lookups at usage time instead of at processing time, i think we're mostly ok here?

sidenote: we really need to start using some proper async primitives (coroutines or flows or _something_) =/

prashanDYDX:
web doesn't need these right? so maybe want to add a config to allow avoid fetching this?

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipChainResourceProcessor.kt
mike-dydx: > at usage sites
meaning client?

> that way if the rpc configs update we don't have an outdated endpoints hanging around
the rpc endpoints are updated only when chains are updated. I think by fixing the race condition Rui pointed out, they will either both be stale or both be up-to-date. Maybe I'm misunderstanding though

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
mike-dydx: > mostly ok here
yeah unless github.com is extreeeeemely slow.

> we really need to start using some proper async primitives
see latest commit https://github.com/dydxprotocol/v4-abacus/pull/512/commits/20ed7a28620821323577db5f188bde85947eea73

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/RpcInfo.kt
mike-dydx: > is this like every known EVM chain?

yes, based on https://github.com/wevm/viem/tree/main/src/chains/definitions, see [thread](https://dydx-team.slack.com/archives/C042V24U87Q/p1720721749832189) if you are curious

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -73,21 +83,57 @@ internal class OnboardingSupervisor(
...
+        val chainsResponse = chainsRequestDeferred.await()
+
+        if (chainsResponse != null) {
+            update(stateMachine.routerChains(chainsResponse), oldState)
+        }
+    }
+
+    private suspend fun retrieveChainRpcEndpoints() = suspendCancellableCoroutine<Unit> { continuation ->
+        val url = "${helper.deploymentUri}/configs/rpc.json"
+        helper.get(url) { _, response, _, _ ->
```
prashanDYDX, line 126: might be easier (and more sustainable for the future) to add a suspending version of `helper.get()` that does the `suspendCancellableCoroutine` stuff for us.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -73,21 +83,57 @@ internal class OnboardingSupervisor(
...
+            if (response != null) {
+                try {
+                    Json.decodeFromString<Map<String, RpcInfo>>(response).let {
+                        RpcConfigs.chainIdToRpcMap = it
+                    }
+                } catch (e: Exception) {
```
prashanDYDX, line 132: detekt will probably complain about this. we should be catching the exceptions we know about explicitly.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -73,21 +83,57 @@ internal class OnboardingSupervisor(
 
     private fun retrieveAssetsFromRouter() {
         if (StatsigConfig.useSkip) {
-            retrieveSkipTransferChains()
+            MainScope().launch {
```
prashanDYDX, line 87: scopes is where things get tricky. a coroutine scope does not only define the lifecycle within which a coroutine will be run, but also a ton of other metadata such as which CoroutineDispatcher will be used to run it. `MainScope` runs on `Dispatchers.Main` (which runs coroutines on the main thread), but since we're updating Abacus state, we ideally want to be on the Abacus thread as defined by the client. 

I think `Dispatchers.Unconfined` is what we want here, since it'll execute the coroutine in the current thread. But generally, Unconfined is considered a bit of a code smell...

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipChainResourceProcessor.kt
prashanDYDX: yeah basically i was suggesting that you don't even need to fix the race condition if you let callers do the lookup themselves

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/RpcInfo.kt
prashanDYDX: i don't think we need anything super complicated here. just defaulting to like the top few popular evm chains should be fine. ETH, OP, Arbitrum, Polygon, Avalanche, Base

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -73,21 +84,61 @@ internal class OnboardingSupervisor(
...
         } else {
             retrieveTransferAssets()
         }
         retrieveCctpChainIds()
     }
 
-    private fun retrieveSkipTransferChains() {
+
+
+    private suspend fun retrieveSkipTransferChains() = coroutineScope {
```
prashanDYDX, line 99: did compilation fail without `coroutineScope{ }` - i don't think this should be necessary since we're already launching in a scope.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -73,21 +84,61 @@ internal class OnboardingSupervisor(
...
+        val url = "${helper.deploymentUri}/configs/rpc.json"
+        helper.get(url) { _, response, _, _ ->
+            if (response != null) {
+                try {
+                    Json.decodeFromString<Map<String, RpcInfo>>(response).let {
+                        RpcConfigs.chainIdToRpcMap = it
+                    }
+                } catch (e: IllegalArgumentException) {
+                    Logger.e { "retrieveChainRpcEndpoints IllegalArgumentException error: $e" }
+                    throw e  // Rethrow if necessary
```
prashanDYDX, line 135: wait we explicitly do not want to rethrow.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
mike-dydx: Screenshot 2024-07-12 at 10 49 16 AM
this is the result w/o

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/RpcInfo.kt
mike-dydx: what is the benefit of a few vs all of them? Other than a slightly smaller json.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
mike-dydx: 894b4f2515986d3c5be0b3cda66234a6917b4328

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipChainResourceProcessor.kt
mike-dydx: i guess i still lean towards doing this in abacus. Otherwise, 
1. consumer context of the rpc config has to understand the availability ordering of the rpc config data
2. more logic on client-side

Eventually we will probably have a similar scenario, so maybe doing it here can serve as an forward example of how to handle concurrent requests/coroutines?

maybe we can chat live about it.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
mike-dydx: please resolve thread if it looks good after 894b4f2515986d3c5be0b3cda66234a6917b4328

mike-dydx:
> web doesn't need these right? so maybe want to add a config to allow avoid fetching this?

good call out. Resolved in 73b4404fc3b852a09be483e3d5958ec0ef32a1ae

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
mike-dydx: i believe the `coroutineScope` bit establishes scope for the async functions, otherwise, I'd have to do like `CoroutineScope(Dispatchers.Unconfined).async { ...` again

the documentation says:
> Creates a CoroutineScope and calls the specified suspend block with this scope. The provided scope inherits its coroutineContext from the outer scope, but overrides the context's Job.

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/Environment.kt
```diff
@@ -505,6 +505,8 @@ class V4Environment(
             )
         }
 
+        private const val DEFAULT_CHAIN_DECIMALS = 18
+        private const val DEFAULT_NON_CHAIN_DECIMALS = 6
```
mike-dydx, line 509: @ruixhuang is there a better name for these constants? Detekt was yelling about this

ruixhuang (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
mike-dydx: done in https://github.com/dydxprotocol/v4-abacus/pull/512/commits/c51a618703a7d48858c9e06bb70527a462400ef3

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
mike-dydx: addressed

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/RpcInfo.kt
prashanDYDX: don't have to do a file read

mike-dydx: fwiw, i wrote a [script](https://github.com/dydxprotocol/v4-web/pull/798) to update the rpc list from that repo, so it's no more/less work to maintain a more comprehensive set of chains

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -73,20 +82,36 @@ internal class OnboardingSupervisor(
 
     private fun retrieveAssetsFromRouter() {
         if (StatsigConfig.useSkip) {
-            retrieveSkipTransferChains()
+            CoroutineScope(Dispatchers.Unconfined).launch {
```
prashanDYDX, line 85: we should be storing this coroutine scope somewhere (and re-using for other coroutine launches).

prashanDYDX: also let's add `SupervisorJob` as another coroutine context element.

PR #515, opened 2024-07-11
Add InternalState Assets

jaredvu:
**Description**
* Add new type `AssetJson` to represent the return from `${deploymentUri}/configs/markets.json`
   * In order to test this [markets.json](https://github.com/dydxprotocol/v4-web/pull/799) should be the deployment target. Without a perfect `markets.json` file, there will be serialization errors 
Screen Shot 2024-07-11 at 3 20 19 PM

* Added `process` functions that are gated behind the `staticTyping` boolean
* Renamed `MarketProcessor` functions to abide by `process` prefix
* Tested on v4-web with local Abacus


src/commonMain/kotlin/exchange.dydx.abacus/processor/assets/AssetProcessor.kt
```diff
@@ -24,6 +39,34 @@ internal class AssetProcessor(parser: ParserProtocol) : BaseProcessor(parser) {
         ),
     )
 
+    override fun process(
```
ruixhuang, line 42: We can add unit test for this now.

ruixhuang (approved):
Landing this and created a ticket for follow-ups:  https://linear.app/dydx/issue/TRCL-3617/add-unit-tests-for-internalstate-assets

PR #516, opened 2024-07-15
TRCL-3605 Static Typing: Subaccount Orders

ruixhuang:
Guarded by feature flag.  Cancellation doesn't work yet, and will address in follow-ups.

Screenshot_1721056416


src/commonMain/kotlin/exchange.dydx.abacus/output/input/TradeInput.kt
```diff
@@ -673,14 +673,19 @@ enum class OrderStatus(val rawValue: String) {
...
+        operator fun invoke(rawValue: String?): OrderStatus? {
             return if (rawValue == "BEST_EFFORT_OPENED") {
                 Pending
             } else {
                 OrderStatus.values().firstOrNull { it.rawValue == rawValue }
             }
         }
     }
+
+    val isFinalized: Boolean
```
aforaleka, line 685: 🙏

src/commonMain/kotlin/indexer/models/IndexerCompositeOrderObject.kt
```diff
@@ -0,0 +1,43 @@
...
+import indexer.codegen.IndexerAPITimeInForce
+import indexer.codegen.IndexerIsoString
+import indexer.codegen.IndexerOrderSide
+import indexer.codegen.IndexerOrderType
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class IndexerCompositeOrderObject(
+
+    // Copied from IndexerOrderResponseObject
```
ruixhuang, line 13: Websocket returns extra fields, so we can't use the codegen model directly

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
```diff
@@ -1264,7 +1293,7 @@ open class TradingStateMachine(
                 val modifiedFills = fills?.toIMutableMap() ?: mutableMapOf()
                 var subaccountFills = fills?.get(subaccountText)
                 if (staticTyping) {
-                    val newFills = internalState.wallet.account.subaccounts?.get(subaccountNumber)?.fills?.toIList()
```
yogurtandjam, line 1267: was this piece of internal state already released? 
just want to make sure we're isolating state changes

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine.kt
```diff
@@ -1264,7 +1293,7 @@ open class TradingStateMachine(
...
+                    val newFills = internalState.wallet.account.subaccounts[subaccountNumber]?.fills?.toIList()
```
yogurtandjam, line 1296: can we safely assume this if subaccount orders state is hidden behind a feature flag?

yogurtandjam: oh, this is behind the FF. sorry!

yogurtandjam (commented):
can we add tests to make sure that this code behaves as expected for `staticTyping` true vs false?

ruixhuang:
> can we add tests to make sure that this code behaves as expected for `staticTyping` true vs false?
> 
> doesn't have to be super heavy - it's probably fine if we just reuse an existing high level test and inject staticTyping true+ false into it and make sure all the outputs are the same.

Yes, you can set static_type = true when creating the state machine object during test.  However, there are some fields missing from Jared's market asset PR, so the tests would fail.  I will address this in the follow-up PR.

prashanDYDX (approved):

PR #517, opened 2024-07-15
Set ignoreUnknownKeys = true for JSON parsing

ruixhuang:
Otherwise error will be thrown if the incoming data contain unknown keys or enums.  We will need to handle unknown keys for forward compatibility. 

Also moving AssetJson from a previous PR to model class folder, and sets the default values for its optional attributes.


src/commonMain/kotlin/exchange.dydx.abacus/protocols/ParserProtocol.kt
```diff
@@ -71,3 +76,45 @@ inline fun <reified T> ParserProtocol.asTypedList(list: Any?): List<T>? {
...
+        return item
+    }
+    val itemString = if (item is Map<*, *>) {
+        item.toJson()
+    } else {
+        asString(item)
+    }
+    return if (itemString != null) {
+        try {
+            val json = Json {
```
prashanDYDX, line 91: I think we can just cache a singleton `Json` instance.

src/commonMain/kotlin/exchange.dydx.abacus/protocols/ParserProtocol.kt
```diff
@@ -71,3 +76,45 @@ inline fun <reified T> ParserProtocol.asTypedList(list: Any?): List<T>? {
         }
     }
 }
+
+inline fun <reified T> ParserProtocol.asTypedObject(item: Any?): T? {
```
prashanDYDX, line 80: Does this need to be an extension on `ParserProtocol`?

I didn't see you had this, I wrote this in my PR:
```
inline fun <reified T> safeDecode(json: String): T? {
    return try {
        Json.decodeFromString<T>(json)
    } catch (e: SerializationException) {
        Logger.e { "Failed to decode: $json" }
        null
    } catch (e: IllegalArgumentException) {
        Logger.e { "Failed to decode: $json" }
        null
    }
}
```

prashanDYDX (approved):

src/commonMain/kotlin/exchange.dydx.abacus/protocols/ParserProtocol.kt
ruixhuang: Since the code uses parser for all parsing needed, I think it's better to keep it consistent.

src/commonMain/kotlin/exchange.dydx.abacus/protocols/ParserProtocol.kt
ruixhuang: Will change in the follow-ups.

PR #519, opened 2024-07-16
update to support alchemy rpc urls

mike-dydx:
some of the rpc URLs in the viem [definition files](https://github.com/wevm/viem/tree/main/src/chains/definitions) are not correct so we will prefer to use private alchemy rpc nodes. This PR updates to read from a newly-formatted rpc.json file

note, clients will have to now inject a new API key for alchemy via `appConfigs.onboardingConfigs.alchemyApiKey`

see corresponding [web PR](https://github.com/dydxprotocol/v4-web/pull/811)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/Configs.kt
```diff
@@ -170,6 +170,7 @@ data class OnboardingConfigs(
     }
 
     var squidVersion: SquidVersion = SquidVersion.V2
+    var alchemyApiKey: String = ""
```
ruixhuang, line 173: Should this be `null`?

yogurtandjam (approved):

ruixhuang (approved):

PR #521, opened 2024-07-16
TRCL-3612 Static Typing: Historical PNL

ruixhuang:
Guarded by staticTyping flag.  Verified on Android app:

Screenshot_1721152996


src/commonMain/kotlin/exchange.dydx.abacus/utils/ParsingHelper.kt
```diff
@@ -219,6 +219,90 @@ internal class ParsingHelper {
             }
         }
 
+        internal fun <T : Any> mergeTyped(
```
ruixhuang, line 222: This is a copy of the original "merge" but with typed object.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -258,40 +259,81 @@ internal class SubaccountSupervisor(
...
-            ),
-        )?.mutable()
-
-        if (historicalPnl != null) {
-            val last = helper.parser.asMap(historicalPnl.lastOrNull())
-            if (helper.parser.asBool(last?.get("calculated")) == true) {
-                historicalPnl.removeLast()
+        if (stateMachine.staticTyping) {
+            val historicalPNLs = stateMachine.internalState.wallet.account.subaccounts[subaccountNumber]?.historicalPNLs
+            // TODO: remove last if calculated
```
ruixhuang, line 264: We don't calculate the last tick from the client for v4 as on v3, so this is just a placeholder for now

src/commonMain/kotlin/exchange.dydx.abacus/processor/base/BaseProcessor.kt
```diff
@@ -175,54 +175,44 @@ internal open class BaseProcessor(val parser: ParserProtocol) {
...
 
-    internal open fun merge(
+    internal fun <T : Any> merge(
         parser: ParserProtocol,
-        existing: List<Any>?,
-        incoming: List<Any>?,
-        timeField: String,
+        existing: List<T>?,
+        incoming: List<T>?,
+        timeField: (T?) -> Instant?,
```
prashanDYDX, line 182: nit an extension function might be more idiomatic here `T?.() -> Instant?`

prashanDYDX (approved):

PR #522, opened 2024-07-16
Updating existing integration tests for static typing

ruixhuang:
Fixing bad data in the test mocks.
Updating the codegen models and changing to the enum cases for parsing.


src/commonMain/kotlin/exchange.dydx.abacus/protocols/ParserProtocol.kt
```diff
@@ -64,10 +65,12 @@ inline fun <reified T> ParserProtocol.asTypedList(list: Any?): List<T>? {
                     }
                     json.decodeFromString<T>(itemString)
                 } catch (e: SerializationException) {
-                    Logger.e { "Failed to parse item: $item" }
+                    val typeClassifier = typeOf<T>().classifier
```
prashanDYDX, line 68: Reflection is slow, let's avoid if possible.

Since `T` is `reified` here, you should be able to use `T::class`

swagger_codegen.sh
```diff
@@ -41,12 +41,33 @@ mv genereated/src/main/kotlin/indexer/codegen/TransferResponseObject_sender.kt \
...
+sed -i '' 's/TAKEPROFITMARKET/TAKE_PROFIT_MARKET/' genereated/src/main/kotlin/indexer/codegen/APIOrderType.kt
+
 # for each of the time in the generated code, run "swagger_update_file.sh <file>"
 find genereated/src/main/kotlin/indexer -type f \
     -exec $CURRENT_DIR/swagger_update_file.sh {} \;
 
-rm -rf $CURRENT_DIR/src/commonMain/kotlin/indexer
-mv genereated/src/main/kotlin/indexer  $CURRENT_DIR/src/commonMain/kotlin
+rm -rf $CURRENT_DIR/src/commonMain/kotlin/indexer/codegen
+mv genereated/src/main/kotlin/indexer/codegen  $CURRENT_DIR/src/commonMain/kotlin/indexer
```
prashanDYDX, line 70: nit, we have a typo in `generated`

prashanDYDX (approved):

PR #524, opened 2024-07-17
transaction error handling: parse query result error for subaccount update

aforaleka:
while most tx errors are broadcast errors with codespace + code, sometimes there can be errors from simulation/query that are not properly formatted / missing those info. to get around it, i'm adding string matching to parse the error message to see if it starts with "Query failed", and then do additional string matching

in particular, add support for better subaccount update failed result handling, so we can show an error such as 
image
instead of 
image during isolated trade/ margin update. this is usually caught by FE validation but just in case it isn't .3.

- this also needs slight updating on FE/web to make sure we're sending in the Error's error message to abacus (https://github.com/dydxprotocol/v4-web/pull/823)

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/app/adaptors/V4TransactionErrors.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/app/adaptors/V4TransactionErrors.kt
@@ -44,7 +44,6 @@ class V4TransactionErrors {
     }
 }
 
-
 // Copied from protocol subaccounts update (https://github.com/dydxprotocol/v4-chain/blob/b2dfda2a4b0ea587691c41ba436b46d0d9987a25/protocol/x/subaccounts/types/update.go#L58)
 enum class SubaccountUpdateFailedResult(val rawValue: String) {
     NewlyUndercollateralized("NewlyUndercollateralized"),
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/app/adaptors/V4TransactionErrors.kt
```diff
@@ -16,9 +16,45 @@ class V4TransactionErrors {
                 } else {
                     null
                 }
+            } else if (message?.startsWith("Query failed") == true) {
```
moo-onthelawn, line 19: nit: can we extract these strings into a constant somewhere?

moo-onthelawn (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/app/adaptors/V4TransactionErrors.kt
```diff
@@ -16,9 +16,45 @@ class V4TransactionErrors {
...
         }
+
+        private fun parseQueryResultErrorFromMessage(message: String): ParsingError {
+            // Workaround: Regex match different query results until protocol can return codespace/code
+            parseSubaccountUpdateError(message)?.let { return it }
+            return ParsingError(ParsingErrorType.BackendError, "Unknown query result error", null)
+        }
+
+        private fun parseSubaccountUpdateError(message: String): ParsingError? {
+            val failedSubaccountUpdateResultPattern = Regex("""Subaccount with id \{[^}]+\} failed with UpdateResult:\s*([A-Za-z]+):""")
```
ruixhuang, line 33: That's quite hacky..   Let's hope Protocol wont change the error message any time soon..

ruixhuang (approved):

PR #528, opened 2024-07-17
chore: add swap venues

yogurtandjam:
context: https://dydx-team.slack.com/archives/C0738MMRX9C/p1721247027232859

Edge case found for small axelar transfers - non osmosis/neutron swap venues can become preferred (in this case, terra). to remediate this (since we don't have a fallback address for terra) we're restricting swap venues for non-cctp withdrawals and deposits.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/utils/Constants.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/utils/Constants.kt
@@ -26,3 +26,13 @@ internal const val GEO_POLLING_DURATION_SECONDS = 10.0
 
 // Autosweep Constants
 internal const val MIN_USDC_AMOUNT_FOR_AUTO_SWEEP = 20000
+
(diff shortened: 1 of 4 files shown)


rosepuppy (approved):

src/commonMain/kotlin/exchange.dydx.abacus/utils/Constants.kt
```diff
@@ -26,3 +26,13 @@ internal const val GEO_POLLING_DURATION_SECONDS = 10.0
 
 // Autosweep Constants
 internal const val MIN_USDC_AMOUNT_FOR_AUTO_SWEEP = 20000
+
+internal val OSMOSIS_SWAP_VENUE = mapOf(
```
ruixhuang, line 30: Nit:  Prefer keeping those constants in OnboardingSupervisor.kt as private vals, since they are only referenced there.

ruixhuang (approved):

PR #530, opened 2024-07-18
chore(analytics): add sizeInput to place order analytics payload

moo-onthelawn:
to distinguish between when the user has input the amount via usdc size vs size (asset).

tested locally on v4-web:

Screenshot 2024-07-18 at 1 22 07 PM


src/commonTest/kotlin/exchange.dydx.abacus/app/helper/TriggerOrderToastGeneratorTests.kt
```diff
@@ -157,6 +157,7 @@ class TriggerOrderToastGeneratorTests {
                         price = 1000.0,
                         triggerPrice = 1500.0,
                         size = 1.0,
+                        sizeInput = null,
```
yogurtandjam, line 160: can we add a sizeInput for one of these values to confirm that it's piped through properly

moo-onthelawn: oh this should always be null, because trigger orders always have a null [sizeInput](https://github.com/dydxprotocol/v4-abacus/pull/530/files#diff-f684fa19a17638d5b6d17dfd61dd715ab96db87920846ac4c92790e9cd0b8506R1183)

yogurtandjam: thank you! if it ends up being an enormous lift i wouldn't worry about it, but it'd probably be worth tagging somewhere in a comment that we want to test this to make sure we dont have regressions later

src/commonTest/kotlin/exchange.dydx.abacus/app/manager/v2/V4TransactionTests.kt
```diff
@@ -156,7 +156,8 @@ class V4TransactionTests : NetworkTests() {
         subaccountSupervisor?.commitPlaceOrder(0, transactionCallback)
         assertTransactionQueueStarted()
         tradeInput(false, "0.02")
-        subaccountSupervisor?.commitPlaceOrder(0, transactionCallback)
+        val placeOrderPayload = subaccountSupervisor?.commitPlaceOrder(0, transactionCallback)
+        assertEquals("size.size", placeOrderPayload?.sizeInput)
```
yogurtandjam, line 160: nice

yogurtandjam (approved):

src/commonTest/kotlin/exchange.dydx.abacus/app/manager/v2/V4TransactionTests.kt
yogurtandjam: wait, is `sizeInput` supposed to equal `size.size`? or do we need to `parser.value` that

yogurtandjam:
general nit: might be better to call the key `sizeInputPath` to be more descriptive of what to expect

PR #531, opened 2024-07-18
TRCL-3615 Static typing: AccountBalances

ruixhuang:
Guarded by feature flag.  Tested on the Android app.


src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -1615,11 +1616,12 @@ data class AccountBalance(
...
             data: Map<String, Any>,
             decimals: Int,
+            internalState: InternalAccountBalanceState?
         ): AccountBalance? {
             Logger.d { "creating Account Balance\n" }
 
-            val denom = parser.asString(data["denom"])
-            val amount = parser.asDecimal(data["amount"])
+            val denom = internalState?.denom ?: parser.asString(data["denom"])
+            val amount = internalState?.amount ?: parser.asDecimal(data["amount"])
```
yogurtandjam, line 1624: since we're performing a fallback, can we have a log if these values go out of sync unexpectedly? 

could be convenient to log any state discrepancies when we generate the output if it's not too much overhead/too noisy etc.

yogurtandjam (approved):

src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
ruixhuang: It would create more complexity to execute both the old and new code paths when the FF is turned on, so I opt to reply on unit tests instead of run-time verification.  Just too much work unfortunately.

yogurtandjam: @ruixhuang got it. thanks!

PR #534, opened 2024-07-20
TRCL-3632 Static typing:  StakingDelegation

ruixhuang:
Guarded by feature flag. Tested on the Android app.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
@@ -10,6 +10,7 @@ import exchange.dydx.abacus.protocols.LocalizerProtocol
 import exchange.dydx.abacus.protocols.ParserProtocol
 import exchange.dydx.abacus.state.internalstate.InternalAccountBalanceState
 import exchange.dydx.abacus.state.internalstate.InternalAccountState
+import exchange.dydx.abacus.state.internalstate.InternalStakingDelegationState
 import exchange.dydx.abacus.state.internalstate.InternalSubaccountState
 import exchange.dydx.abacus.state.manager.TokenInfo
 import exchange.dydx.abacus.utils.IList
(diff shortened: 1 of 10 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
```diff
@@ -2218,22 +2220,37 @@ data class Account(
...
+                processStakingBalanceDeprecated(
+                    existing = existing,
+                    parser = parser,
+                    data = data,
+                    tokensInfo = tokensInfo,
+                )
+            }
 
-            val stakingDelegations: IMutableList<StakingDelegation> =
+            val stakingDelegations = if (staticTyping) {
```
yogurtandjam, line 2239: [discussion] the scale of the refactor is pretty enormous. do you think it makes sense to use a separate variable for each piece of state? this would allow us to isolate changes really effectively and be pretty cautious about roll out
cc: @prashanDYDX

ruixhuang: There is a separate field for each piece of state already.  I will update the params to make it more clear.

The calculation logic operates on multiple fields from the state, so it will be difficult to isolate the changes (e.g., using separate flags).

src/commonMain/kotlin/indexer/models/chain/OnChainDelegationResponse.kt
```diff
@@ -0,0 +1,23 @@
+package indexer.models.chain
+
+import kotlinx.serialization.Serializable
+
+@Serializable
```
yogurtandjam, line 5: [q] do objects not natively come with `.toString()` in abacus? what does `@Serializable` actually do?

yogurtandjam (approved):

src/commonMain/kotlin/exchange.dydx.abacus/output/Account.kt
yogurtandjam: makes sense to me!

src/commonMain/kotlin/indexer/models/chain/OnChainDelegationResponse.kt
ruixhuang: @Serializable is needed to decode the json string to the data class.

PR #537, opened 2024-07-23
TRCL-3621 Static typing: Account: PerpetualPositionsProcessor

ruixhuang:
There are calculated fields in the openPositions still to be converted, and we will get to them when working on SubaccountCalculator.  

Tested on Android app.


src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/PerpetualPositionsProcessor.kt
```diff
@@ -1,13 +1,71 @@
...
+        } else {
+            existing
+        }
+    }
+
+    fun processChanges(
+        existing: Map<String, InternalPerpetualPosition>?,
+        payload: List<IndexerPerpetualPositionResponseObject>?,
+    ): Map<String, InternalPerpetualPosition>? {
+        return if (payload != null) {
```
yogurtandjam, line 48: [non-blocking] easier for me to read if we use a guard clause 🙏 . also, wouldn't 
```
            return if (result != existing) {
                result
            } else {
                existing
            }
```
be the same as
```
return result
```
since if result is == to existing, it's the same thing as returning existing? or is there a referential equality check somewhere that we need to fulfill?

```suggestion
        if (payload == null) return existing
        val result = existing?.toMutableMap() ?: mutableMapOf()
        for (item in payload) {
                if (item.market != null) {
                    val newPosition = itemProcessor.processChanges(result[item.market], item)
                    if (newPosition != null) {
                        result[item.market] = newPosition
                    } else {
                        result.remove(item.market)
                    }
                }
            }
            return result
```

yogurtandjam (approved):

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/PerpetualPositionsProcessor.kt
ruixhuang: It's copied from the old code, but I don't think it's really needed at the processor level.  I will update them in follow-up PRs.

PR #539, opened 2024-07-25
TRCL-3646: Static typing: StakingRewards and StakingUnbonding

ruixhuang:


src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AccountProcessor.kt
```diff
@@ -199,15 +171,14 @@ internal class V4AccountProcessor(
...
 
     internal fun processAccountBalances(
         existing: InternalAccountState,
         payload: List<OnChainAccountBalanceObject>?,
     ): InternalAccountState {
-        val balances = balancesProcessor.process(existing.balances, payload)
-        if (balances != existing.balances) {
-            existing.balances = balances
-        }
+        existing.balances = balancesProcessor.process(existing.balances, payload)
```
yogurtandjam, line 181: 👍

yogurtandjam: [suggestion] honestly this is so clean now i wonder if we even need a function for this? could just be an inline call.

if we do decide to do it, am in favor of doing it in a followup PR

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+Wallet.kt
```diff
@@ -296,24 +298,46 @@ internal fun TradingStateMachine.onChainDelegations(payload: String): StateChang
...
-    val response = parser.decodeJsonObject(payload)
-    return try {
-        this.wallet = walletProcessor.receivedStakingRewards(wallet, response)
-        return StateChanges(iListOf(Changes.accountBalances), null)
-    } catch (e: Exception) {
-        StateChanges(iListOf())
+    if (staticTyping) {
+        val response = parser.asTypedObject<OnChainStakingRewardsResponse>(payload)
+        val oldValue = internalState.wallet.account.stakingRewards
+        walletProcessor.processStakingRewards(internalState.wallet, response)
```
yogurtandjam, line 328: [q] does `oldValue` not get updated when we mutate the internal state via the `process` methods?

ruixhuang: Yeah, oldValue contains a copy of the old state, and then we can compare and set the "changes set" only if there is a change.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AccountProcessor.kt
ruixhuang: We can potentially reduce the hierarchy of the processors in the future, but for now I think it's better to keep the existing hierarchy so that we have a one-to-one mapping between the old parsers and new ones.

Also, it can be argued that although the hierarchy of processors creates lots of boilerplate code, it makes the code better structured (and hopefully easier to maintain once you understand the structure)

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AccountProcessor.kt
```diff
@@ -199,15 +171,14 @@ internal class V4AccountProcessor(
     private val delegationsProcessor = AccountDelegationsProcessor(parser)
     private val tradingRewardsProcessor = AccountTradingRewardsProcessor(parser)
     private val launchIncentivePointsProcessor = LaunchIncentivePointsProcessor(parser)
+    private val stakingRewardsProcessor = StakingRewardsProcessor(parser)
```
rosepuppy, line 174: Thanks for breaking these out!

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AccountProcessor.kt
yogurtandjam: yeah 100% agree on not doing this now, if we even decide to do it.

fair, i think that's probably a bigger discussion to have! thanks rui

yogurtandjam (approved):
LGTM but may also want bill's approval

PR #541, opened 2024-07-25
TRCL-3620 Static typing: Subaccount transfers

ruixhuang:
Tested on Android.  Guarded by feature flag.


src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AccountProcessor.kt
```diff
@@ -327,14 +328,25 @@ internal class V4AccountProcessor(
         return modified
     }
 
-    internal fun receivedTransfers(
+    fun processTransfers(
+        existing: InternalAccountState,
+        payload: List<IndexerTransferResponseObject>?,
```
prashanDYDX, line 333: can we just skip this call entirely if null?

ruixhuang: We can.. but the subaccount processor already handles it for us.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
```diff
@@ -130,6 +131,22 @@ internal open class SubaccountProcessor(
             payload = positions,
         )
 
+        val transfers: List<IndexerTransferResponseObject>?
+        val transferList = parser.asTypedList<IndexerTransferResponseObject>(content["transfers"])
+        if (transferList != null) {
+            transfers = transferList
+        } else {
+            val transfer = parser.asTypedObject<IndexerTransferResponseObject>(content["transfers"])
+            transfers = if (transfer != null) listOf(transfer) else null
```
yogurtandjam, line 140: [q] wait, `content["transfers"]` can either a list of transfers for a singel transfer object? 👀 so weird

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
```diff
@@ -357,14 +371,11 @@ internal open class SubaccountProcessor(
         payload: List<IndexerFillResponseObject>?,
         reset: Boolean,
     ): InternalSubaccountState {
-        val newFills = fillsProcessor.process(
+        subaccount.fills = fillsProcessor.process(
```
yogurtandjam, line 374: [discussion] definitely not for this PR, but i imagine these Processor classes don't really need to be classes, since they're just consuming an input and returning an digested output? cc @prashanDYDX do we have plans of converting these to functions?

yogurtandjam (approved):

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: Yeah.. the initial call potentially returns a list , and then the subsequent update is a single object. 
 There are integration tests depending on those.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: IMHO, having a class instead of just static function for processor actually makes unit testing easier when there are delegations.  Prefer keeping it as-is, but we can revisit this when we have time.

PR #543, opened 2024-07-25
feat: Add svm support to skip processor 

tian000:
Adds SVM support to skip API calls and the skip route processor.

Chains and assets returned by skip are filtered by the apps, so Solana chains and assets should appear hidden until filters on the apps are modified to include them.


yogurtandjam (approved):
LGTM!

PR #544, opened 2024-07-25
remove transfer receipt lines which have unavailable data

mike-dydx:
exchange rate, exchange received, and gas fee are not available from Skip, so i have conditionally removed

| Before | After |
| ------ | ----- |
|  |  |


prashanDYDX (approved):

prashanDYDX:
is there a way to test this easily?

PR #546, opened 2024-07-26
TRCL-3647 Static-typing: AssetPositions (QuoteBalance)

ruixhuang:
Fetching AssetPositions and calculating QuoteBalance.  Guarded by feature flag.


src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AssetPositionsProcessor.kt
```diff
@@ -2,18 +2,60 @@ package exchange.dydx.abacus.processor.wallet.account
...
 
-internal class AssetPositionsProcessor(parser: ParserProtocol) : BaseProcessor(parser) {
-    private val itemProcessor = AssetPositionProcessor(parser = parser)
+internal class AssetPositionsProcessor(
+    parser: ParserProtocol,
+    private val itemProcessor: AssetPositionProcessorProtocol = AssetPositionProcessor(parser = parser)
+) : BaseProcessor(parser) {
 
-    internal fun received(payload: Map<String, Any>?): Map<String, Any>? {
+    // From REST call
```
yogurtandjam, line 15: [suggestion] since we have the opportunity to make new names, can we make them descriptive of the source that's calling them? 
e.g.:
processRESTCall

vs

processWebsocketMessage

or something

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
```diff
@@ -107,9 +110,16 @@ internal open class SubaccountProcessor(
         state = process(
             existing = state,
             payload = subaccount,
-            firstTime = true,
+            firstTime = subscribed,
```
yogurtandjam, line 113: [q] does this mean after the first time processing, we will be setting the `subscribed` bool to false?

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AssetPositionsProcessor.kt
ruixhuang: Yeah.. Let's do the naming change afterwards...  There are some complexities involving the current tests code (where it processes the REST function whereas in the real app the data is actually from Websocket) that we will have to clean up.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: Right... the same process call is used for both cases.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AssetPositionsProcessor.kt
yogurtandjam: got it, sounds good! can we add a ticket to clean up and put as a comment?

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
yogurtandjam: wait, the same process call is used for both cases? both cases as in the first time and second time? not 100% sure what you mean in terms of that

yogurtandjam (approved):

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: Yes, but the first time and the second time.

PR #548, opened 2024-07-29
Static Typing: UpdateHeight and OrderCancelled

ruixhuang:
Guarded by feature flag.  Tested on Android.


src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/WalletProcessor.kt
```diff
@@ -121,14 +121,31 @@ internal class WalletProcessor(
         }
     }
 
-    internal fun updateHeight(
+    fun updateHeight(
+        existing: InternalWalletState?,
+        height: BlockAndTime?,
+    ): Triple<InternalWalletState?, Boolean, List<Int>?> {
```
prashanDYDX, line 127: should we just make a data type for this? the `Boolean` in particular is not super clear

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/AccountProcessor.kt
```diff
@@ -685,15 +703,29 @@ internal class V4AccountProcessor(
...
     }
 
-    internal fun orderCanceled(
+    fun orderCanceled(
+        existing: InternalAccountState,
+        orderId: String,
+        subaccountNumber: Int,
+    ): Pair<InternalAccountState, Boolean> {
+        val subaccounts = existing.subaccounts
+        val (modifiedSubaccounts, updated) = subaccountsProcessor.orderCanceled(
```
prashanDYDX, line 712: we don't use `modifiedSubaccounts` looks like

prashanDYDX (approved):

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/WalletProcessor.kt
ruixhuang: Yeah.. we can do that later... right now it's just using the same return type as the old code.

PR #549, opened 2024-07-29
TRCL-3613 Static Typing: TradingRewards

ruixhuang:
Guarded by feature flag.  Tested it on Android


src/commonMain/kotlin/exchange.dydx.abacus/output/account/TradingRewards.kt
```diff
@@ -36,6 +39,56 @@ data class TradingRewards(
 ) {
     companion object {
         internal fun create(
+            existing: TradingRewards?,
```
prashanDYDX, line 42: we shouldn't need to pass in `existing` here, we can just grab it from the `internalState` right?

prashanDYDX (approved):

src/commonMain/kotlin/exchange.dydx.abacus/output/account/TradingRewards.kt
ruixhuang: `internalState` contains the indexer data in this case, and `existing` is the output data type..  The data come from multiple sources processed at different processors, so it's easier to do the data aggregation at the output.

PR #552, opened 2024-07-31
Refacor SubaccountSupervisor

ruixhuang:
`SubaccountSupervisor` is doing too much.  Breaking it apart with the following:

`SubaccountTransactionPayloadProvider`: Producing the payloads for various transactions
`SubaccountTransactionSupervisor`: Submitting transactions
`SubaccountTransactionTracker`: Recording analytics events for transactions

Just moving the same code around... no logic change.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
@@ -1,20 +1,6 @@
 package exchange.dydx.abacus.state.v2.supervisor
 
-import abs
-import exchange.dydx.abacus.calculator.MarginCalculator
-import exchange.dydx.abacus.calculator.TriggerOrdersConstants.TRIGGER_ORDER_DEFAULT_DURATION_DAYS
 import exchange.dydx.abacus.output.Notification
-import exchange.dydx.abacus.output.account.PositionSide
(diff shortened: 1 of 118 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
```diff
@@ -95,6 +64,26 @@ internal class SubaccountSupervisor(
...
     internal val subaccountNumber: Int,
 ) : DynamicNetworkSupervisor(stateMachine, helper, analyticsUtils) {
+
+    private val payloadProvider: SubaccountTransactionPayloadProviderProtocol = SubaccountTransactionPayloadProvider(
+        stateMachine = stateMachine,
+        subaccountNumber = subaccountNumber,
+        helper = helper,
+        accountAddress = accountAddress,
+    )
+    private val transactionSupervisor = SubaccountTransactionSupervisor(
```
prashanDYDX, line 74: probably want to pass into constructor for better testability

prashanDYDX (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountSupervisor.kt
ruixhuang: Yes, we can do that when adding unit tests to the class.

PR #556, opened 2024-08-02
TRCL-3652 Static Typing: MarketProcessor

ruixhuang:
Tested on Android.  Guarded by feature flag.


src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/MarketsProcessor.kt
```diff
@@ -1,25 +1,109 @@
...
+    private val marketProcessor: MarketProcessorProtocol = MarketProcessor(parser, calculateSparklines)
+) : BaseProcessor(parser), MarketsProcessorProtocol {
+    private val marketProcessorDeprecated: MarketProcessor? = marketProcessor as? MarketProcessor
 
     internal var groupingMultiplier: Int
-        get() = marketProcessor.groupingMultiplier
+        get() = marketProcessorDeprecated!!.groupingMultiplier
         set(value) {
-            marketProcessor.groupingMultiplier = value
+            marketProcessorDeprecated?.groupingMultiplier = value
```
jaredvu, line 41: why the `?.` safe accessor here but the `!!.` in the `get`?

src/commonMain/kotlin/indexer/models/configs/ConfigsMarketAssetResponse.kt
jaredvu, line 1: Not sure this belongs under `indexer/models` since it doesn't come from `indexer`

jaredvu (approved):
Overall looks great!

src/commonMain/kotlin/indexer/models/configs/ConfigsMarketAssetResponse.kt
ruixhuang: Good point... Let's rename the it something like "responses/models" ?  We can do this in a later PR

src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/MarketsProcessor.kt
ruixhuang: Either is fine.... it's there just to keep the legacy code happy.

PR #558, opened 2024-08-06
TRCL-3606 Static typing: Orderbook

ruixhuang:
The post-fetch calculation logic is now moved to a separate class. 


src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/OrderbookProcessor.kt
```diff
@@ -19,7 +33,101 @@ internal class OrderbookProcessor(parser: ParserProtocol) : BaseProcessor(parser
...
+                changes = content?.bids,
+                ascending = false,
+            ),
+        )
+        existing.consolidatedOrderbook = calculator.consolidate(
+            rawOrderbook = existing.rawOrderbook,
+        )
+        existing.groupedOrderbook = calculator.calculate(
+            rawOrderbook = existing.rawOrderbook,
+            tickSize = tickSize ?: 0.1,
```
jaredvu, line 107: we use `0.1` as a default tickSize. Would it be useful to store as a const to reduce amount of magic numbers in this processor?

ruixhuang: 👍  Addressed.

jaredvu (approved):

PR #559, opened 2024-08-06
TRCL-3658 Static typing: Market Candles

ruixhuang:

Also updated the codegen script to ensure correct parsing of the CandleResolutions.


src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/CandlesProcessor.kt
```diff
@@ -2,12 +2,97 @@ package exchange.dydx.abacus.processor.markets
...
+        return existing
+    }
+
+    fun processBatchUpdate(
+        existing: InternalMarketState,
+        resolution: String,
+        payload: List<IndexerCandleResponseObject>?
+    ): InternalMarketState {
+        if (!payload.isNullOrEmpty()) {
+            val candles = payload.reversed().mapNotNull {
```
jaredvu, line 51: why does the payload need to be reversed? Gut feeling is that the candle merge fn should be agnostic of the order and responsible for the sort instead of sorting payload first

src/commonMain/kotlin/exchange.dydx.abacus/state/internalstate/InternalState.kt
```diff
@@ -57,6 +58,9 @@ internal data class InternalMarketState(
 
     // grouped orderbook for app display
     var groupedOrderbook: MarketOrderbook? = null,
+
+    // candles: resolution -> candles
+    var candles: MutableMap<String, List<MarketCandle>>? = null
```
jaredvu, line 63: Should this be `MutableMap<String, MutableMap<String, List<MarketCandle>>>`? Market -> Resolution -> Candles

src/commonMain/kotlin/exchange.dydx.abacus/processor/markets/CandlesProcessor.kt
ruixhuang: merge() assumes in inputs are sorted, which makes sense since this way it wouldn't need to call sort() every time.  It is more efficient with this assumption.

src/commonMain/kotlin/exchange.dydx.abacus/state/internalstate/InternalState.kt
ruixhuang: The parent object is keyed by "MarketId".

PR #561, opened 2024-08-07
fix: geo poll less to avoid cloudflare rate limiting

tyleroooo:


aforaleka (approved):

rosepuppy (approved):

src/commonMain/kotlin/exchange.dydx.abacus/utils/Constants.kt
```diff
@@ -24,7 +24,7 @@ internal const val ETHEREUM_CHAIN_ID = "1"
 internal val ALLOWED_CHAIN_TYPES = listOf("evm", "svm")
 
 // Polling durations
-internal const val GEO_POLLING_DURATION_SECONDS = 10.0
+internal const val GEO_POLLING_DURATION_SECONDS = 180.0
```
rosepuppy, line 27: comfortable changing this to 300.0 as well

rosepuppy:
We decided to just get rid of the polling completely, I will put up another PR

PR #567, opened 2024-08-09
Allow cosmos chains deposit/withdrawls if Keplr is the connected wallet [OTE-685]

rosepuppy:
We want to include cosmos chains for onboarding when using keplr wallet

Screenshot 2024-08-09 at 11 02 23 AM


src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
```diff
@@ -27,6 +28,13 @@ internal class SkipProcessor(
 
     var skipTokens: Map<String, Map<String, List<Map<String, Any>>>>? = null
     override var exchangeDestinationChainId: String? = null
+    override var cosmosWalletConnected: Boolean? = false
```
yogurtandjam, line 31: [suggestion] we'll have phantom wallet integrated soon enough, can we create an enum to describe all the possible wallets instead? then we can do something like `ALOWED_CHAIN_TYPES + SELECTED_WALLET_CHAIN_TYPE` on line 306 to make this easily extendable to new wallets

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
```diff
@@ -294,11 +302,16 @@ internal class SkipProcessor(
...
+            ALLOWED_CHAIN_TYPES + COSMOS_CHAIN_TYPES
+        } else {
+            ALLOWED_CHAIN_TYPES
+        }
 
         this.chains?.let {
             for (chain in it) {
                 parser.asNativeMap(chain)?.let { chain ->
-                    if (parser.asString(chain.get("chain_type")) in ALLOWED_CHAIN_TYPES) {
+                    if (parser.asString(chain.get("chain_type")) in allowChainTypes) {
```
yogurtandjam, line 314:
```suggestion
                    if (parser.asString(chain.get("chain_type")) in SELECTED_WALLET_CHAIN_TYPE) {
```

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
rosepuppy: I can take a stab

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
```diff
@@ -27,6 +27,13 @@ internal class SkipProcessor(
 
     var skipTokens: Map<String, Map<String, List<Map<String, Any>>>>? = null
     override var exchangeDestinationChainId: String? = null
+    override var selectedChainType: ChainType? = ChainType.EVM
+        set(value) {
+            if (field != value) {
+                field = value
+                internalState.chains = chainOptions()
```
yogurtandjam, line 34: will this force the abacus state to update? since we're not triggering a a state update for updating the `selectedChainType`

we might want to make this a piece of state so the recalculation happens automatically

rosepuppy: It does seem like Rui recently updated so that we use `internalState.chains` directly for the state output:
https://github.com/dydxprotocol/v4-abacus/blob/main/src/commonMain/kotlin/exchange.dydx.abacus/output/input/TransferInput.kt#L44

yogurtandjam (dismissed):

yogurtandjam (approved):

PR #572, opened 2024-08-14
feat: evm swaps for deposits [OTE-700]

yogurtandjam:
enables evm swaps by:
- calling the swap venues endpoint and storing the uniswap venues internally in the class. i opted to keep the list as a class variable instead of throwing it into state for the following reasons:
  - it doesn't need to be updated 
  - it doesn't need to be exposed to the clients
  - considering the previous two, feels like a whole lot of boilerplate/overhead for no material gain
  - it should only be referenced specifically for calling skip endpoints and not be available to parts of the code that do not directly work skip endpoints
  - when @ruixhuang 's work to migrate everything to `internalState` is done, we can really easily move it into that state.
- adding the `swap_venues` list to the nonCCTP deposit payload (this enables the request to to include evm swap venues)
- adding `CCTP` to the list of enabled bridges for nonCCTP deposits
  - evm swaps often rely on `CCTP` bridges
  - this means that CCTP enabled chains will route via CCTP even if we don't explicitly mark them as such. fortunately there's no chains that fit this criteria and we're exploring refactorings as a direct follow up to this PR to make sure this doesn't happen (or to unify the 2 paths, if there's no reason to keep the 2 paths separate anymore)

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
@@ -16,7 +16,6 @@ import exchange.dydx.abacus.utils.safeSet
 
 private const val UNISWAP_SUFFIX = "uniswap"
 
-@Suppress("NotImplementedDeclaration")
 internal class SkipProcessor(
     parser: ParserProtocol,
     private val internalState: InternalTransferInputState
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
@@ -27,8 +27,6 @@ internal class SkipProcessor(
 //    actual type of the tokens payload is Map<str, Map<str, List<Map<str, Any>>>>
     override var tokens: List<Any>? = null
 
-    override var evmSwapVenues: List<Any?> = listOf()
-
     var skipTokens: Map<String, Map<String, List<Map<String, Any>>>>? = null
     override var exchangeDestinationChainId: String? = null
 
@@ -79,7 +77,7 @@ internal class SkipProcessor(
             )
         }
         if (evmSwapVenues != null) {
-            this.evmSwapVenues = evmSwapVenues
+            this.internalState.evmSwapVenues = evmSwapVenues
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
```diff
@@ -60,6 +63,25 @@ internal class SkipProcessor(
...
+        payload: Map<String, Any>
+    ) {
+        val venues = parser.asNativeList(payload.get("venues"))
+        val evmSwapVenues = venues?.filter {
+            parser.asString(parser.asMap(it)?.get("name"))?.endsWith(UNISWAP_SUFFIX) == true
+        }?.map {
+            val swapVenue = parser.asMap(it)
+            mapOf(
+                "name" to parser.asString(swapVenue?.get("name")),
+                "chain_id" to parser.asString(swapVenue?.get("chain_id")),
```
yogurtandjam, line 77: am aware that in a different time the recommended pattern would be to create a processor. it just feels too heavy weight fro something this simple

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -131,11 +137,19 @@ internal class OnboardingSupervisor(
         }
     }
 
+    private fun retrieveSkipEvmSwapVenues() {
```
yogurtandjam, line 140: we can alternatively do this by making it async and calling it in each routing call to make sure the swap venues info is most up to date. it feels a little excessive to do that but im open to opinions here

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt
```diff
@@ -14,7 +14,8 @@ import exchange.dydx.abacus.utils.NATIVE_TOKEN_DEFAULT_ADDRESS
 import exchange.dydx.abacus.utils.mutable
 import exchange.dydx.abacus.utils.safeSet
 
-@Suppress("NotImplementedDeclaration", "ForbiddenComment")
+private const val UNISWAP_SUFFIX = "uniswap"
```
rosepuppy, line 17: for my understanding is uniswap the only dex supported?

yogurtandjam: yup! right now we're just hardcoding it but i'm considering moving to statsig 🤔 so we can be more nimble.

rosepuppy (dismissed):
🚢

prashanDYDX (dismissed):

prashanDYDX (approved):

PR #574, opened 2024-08-15
[chore] ff evm swaps

yogurtandjam:
adds a feature flag to gate evm swap functionality

all options altered by evm swap functionality are defined in separate objects and gated by evm swap flag

gates the addition of evm swap venues to state behind feature flag. this could be done more efficiently by gating the API call itself

### DIFF
--- a/v4_abacus.podspec
+++ b/v4_abacus.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'v4_abacus'
-    spec.version = '1.8.89'
+    spec.version                  = '1.8.89'
     spec.homepage                 = 'https://github.com/dydxprotocol/v4-abacus'
     spec.source                   = { :http=> ''}
     spec.authors                  = ''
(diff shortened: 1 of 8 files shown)


prashanDYDX (dismissed):

v4_abacus.podspec
```diff
@@ -11,7 +11,7 @@ Pod::Spec.new do |spec|
                 
                 
                 
-    if false
+    if !Dir.exist?('build/cocoapods/framework/Abacus.framework') || Dir.empty?('build/cocoapods/framework/Abacus.framework')
```
ruixhuang, line 14: Can we revert this?  It causes issue with iOS build

yogurtandjam: wait, if we just have `if false` as the condition, can't we just delete this whole block? seems like it should never hit

prashanDYDX: this file is auto generated, it's simpler to just replace the condition with false

prashanDYDX (approved):

PR #578, opened 2024-08-19
feat: vaults helper functions

tyleroooo:


src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/Vault.kt
```diff
@@ -0,0 +1,274 @@
...
+import indexer.codegen.IndexerPerpetualPositionResponseObject
+import kotlinx.datetime.Clock
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+import kotlin.time.Duration.Companion.days
+import kotlin.time.Duration.Companion.milliseconds
+
+@JsExport
+@Serializable
+data class IndexerHistoricalPnl(
```
prashanDYDX, line 23: I think it might be easier for others to read in the future if we move these data models to another file

tyleroooo: Hmm do you have an opinion on what to pull out? All types? Separate input and output types for each file? Each function? Should we include the basic parsing functions?

ruixhuang: Model classes can be moved to `src/commonMain/indexer/models/`.  We can created a subfolder "vault" there.  It's fine to keep them in a single file.

The top-level functions (`calculateVaultPositions` for example) should probably live in a separate file.  I'd prefer moving them into a class (`VaultCalculator`?) so that it's dependencies (`SubaccountCalculatorV2`, etc) can be supplied into its constructor.  This is a common pattern to simplify unit tests.

prashanDYDX: The idea behind moving things into a class, and specifically moving dependencies into the constructor is for dependency inversion.

This gives us the ability to easily swap pass in fake implementations at test time. And, down the line if we choose to move to full dependency injection, will make our lives much easier so we don't have to worry about passing dependencies everywhere.

I agree that we shouldn't just blindly apply patterns where they may not be helpful, and in this specific case may not provide a ton of value since testing the matrix of inputs and outputs of the entire system is relatively straightforward. The value would mostly be in establishing patterns that would help developers in the future.

I think minimally, we should move the functions into an `object VaultCalculator` (or similar) just to clean up the global namespace.

tyleroooo: Okay, moved to object Calculators (not classes for now). I left the data classes in the global scope as seems to be the norm.

ruixhuang (approved):

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultIndexerModels.kt
jaredvu, line 1: Is the new pattern going to be co-locating the models within the `functional/${feature_name}` directories?

jaredvu: If it is coming from Indexer will this eventually be moved to the `/indexer/codegen`?

jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultIndexerModels.kt
tyleroooo: yep, this is just a temporary file, to be swagger generated as soon as it's available.

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/Vault.kt
```diff
@@ -0,0 +1,231 @@
+package exchange.dydx.abacus.functional.vault
```
yogurtandjam, line 1: 🔥 🔥 FUNCTIONAL 🔥 🔥

src/commonTest/kotlin/exchange.dydx.abacus/functional/vault/VaultTests.kt
```diff
@@ -0,0 +1,205 @@
+package exchange.dydx.abacus.functional.vault
```
yogurtandjam, line 1: 🔥 🔥 🔥 UNIT TESTS 🔥 🔥 🔥

yogurtandjam (approved):
i just wanted to stamp again bc of how happy this PR makes me

yogurtandjam (approved):

PR #581, opened 2024-08-20
chore: datadog poc [OTE-719][1/n]

yogurtandjam:
This PR:
- updates Logger.kt's `e` function to accept a context (IMap so its json serializable), and error (i'm using Kotlin's default Error which is probably the wrong type, still investigating this one)
- adds `ddInfo` to Logger.kt. This behaves similarly to `Logger.d` except that it ships logs to datadog and accepts a context IMap

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/utils/Logger.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/utils/Logger.kt
@@ -17,10 +17,18 @@ object Logger {
         }
     }
 
-    fun e(message: () -> String) {
+    fun e(context: Map<String, Any>? = null, error: Error? = null, message: () -> String) {
         clientLogger?.let {
-            it.e(TAG, message())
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/utils/Logger.kt
```diff
@@ -17,11 +17,19 @@ object Logger {
         }
     }
 
-    fun e(message: () -> String) {
+    fun e(message: () -> String, context: IMap<String, Any>? = null, error: Error? = null) {
```
ruixhuang, line 20: Moving `message` as the last param.  This way, you don't have to change existing calls.

yogurtandjam: sounds good to me!

ruixhuang (approved):

tyleroooo (approved):

PR #585, opened 2024-08-21
feat(isolatedMargin): add in max leverage calculation

moo-onthelawn:
## Goal
Cap the leverage slider at user's max leverage for that market based on their current position size.

## Implementation
Replace the calculation on `TradeInputOptions.maxLeverage` which is currently defined as `max leverage for the leverage slider` [src](https://github.com/dydxprotocol/v4-abacus/blob/main/docs/Input/TradeInput.md#maxleverage).

## Notes
- Mobile concerns: As far as I can tell, in web, we don't use `TradeInputOptions.maxLeverage` anywhere lmao so this is safe (I just have to make changes to actually use this value in our slider). Unsure about mobile, so gonna review with mobile team to ensure I'm not breaking anything; but ideally if mobile is using this property as intended, it's just free.
- This calculation is an approximation and sometimes leaves the user with money leftover (i.e. they could theoretically put in a larger size), and sometimes is slightly over 100% margin (depending on the state of the orderbook)
  - Can file a follow up to make it more accurate but it's a non trivial amount of work to loop through the orderbook so would rather not do it right now 🫠 

## Prototype onto v4-web
*Ignore how buggy our slider is ugh; dragging it is a journey. Will try to do some work on FE code to fix that. Main point is that the max's are generally accurate
- lol jk my vids aren't uploading, will try in slack

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
@@ -61,7 +61,7 @@ internal class TradeInputCalculator(
         val account = parser.asNativeMap(state["account"])
 
         val crossMarginSubaccount = parser.asNativeMap(parser.value(account, "subaccounts.$subaccountNumber"))
-        val groupedSubaccounts = parser.asMap(parser.value(account, "groupedSubaccounts.$subaccountNumber"))
+        val groupedSubaccount = parser.asMap(parser.value(account, "groupedSubaccounts.$subaccountNumber"))
             ?: crossMarginSubaccount
 
         val user = parser.asNativeMap(state["user"]) ?: mapOf()
@@ -78,8 +78,9 @@ internal class TradeInputCalculator(
         val subaccount = if (marginMode == MarginMode.Cross) {
             crossMarginSubaccount
         } else {
-            groupedSubaccounts
-        } // TODO: incorrect for isolated trades; fix CT-1092
+            // TODO: incorrect for isolated trades; fix CT-1092
+            groupedSubaccount
+        }
 
         val marketId = parser.asString(trade?.get("marketId"))
         val type = parser.asString(trade?.get("type"))
@@ -855,22 +856,11 @@ internal class TradeInputCalculator(
         feeTiers: List<Any>?,
         type: String,
     ): Map<String, Any> {
-        val marketId = parser.asString(market?.get("id"))
-        val position = if (marketId != null) {
-            parser.asNativeMap(
-                parser.value(
-                    subaccount,
-                    "openPositions.$marketId",
-                ),
-            )
-        } else {
-            null
-        }
         var modified = trade.mutable()
         val fields = requiredFields(account, subaccount, trade, market)
         modified.safeSet("fields", fields)
-        modified.safeSet("options", calculatedOptionsFromFields(fields, trade, position, market, subaccount))
-        modified = defaultOptions(account, subaccount, modified, position, market)
+        modified.safeSet("options", calculatedOptionsFromFields(fields, trade, market, subaccount))
+        modified = defaultOptions(account, subaccount, modified, market)
         modified.safeSet(
             "summary",
             summaryForType(trade, subaccount, user, market, rewardsParams, feeTiers, type),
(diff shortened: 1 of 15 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputOptionsCalculator.kt
```diff
@@ -340,17 +343,27 @@ internal class TradeInputOptionsCalculator(
...
+        val equity = subaccount.calculated[CalculationPeriod.current]?.equity
+        val freeCollateral = subaccount.calculated[CalculationPeriod.current]?.freeCollateral
+        val initialMarginFraction =
+            market.perpetualMarket?.configs?.effectiveInitialMarginFraction ?: return null
+
+        val maxMarketLeverage = Numeric.double.ONE / initialMarginFraction
+        val positionNotionalTotal = position?.calculated?.get(CalculationPeriod.current)?.notionalTotal ?: Numeric.double.ZERO
+
+        return if (equity != null && freeCollateral != null) {
+            (freeCollateral + positionNotionalTotal / maxMarketLeverage) * maxMarketLeverage / equity
```
ruixhuang, line 364: Can we check to ensure equity != 0?

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
```diff
@@ -1718,7 +1688,20 @@ internal class TradeInputCalculator(
...
             ),
         )
-        return maxLeverageFromPosition(position, market)
+        val equity = parser.asDouble(parser.value(subaccount, "equity.current"))
+        val freeCollateral = parser.asDouble(parser.value(subaccount, "freeCollateral.current"))
+
+        val initialMarginFraction =
+            parser.asDouble(parser.value(market, "configs.effectiveInitialMarginFraction"))
+                ?: return null
+        val maxMarketLeverage = Numeric.double.ONE / initialMarginFraction
```
mike-dydx, line 1697: doubt it'd happen, but for safety's sake
```suggestion
        val maxMarketLeverage = if (initialMarginFraction <= 0.0) {
    return null
} else {
    Numeric.double.ONE / initialMarginFraction
}
```

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
```diff
@@ -1718,7 +1688,20 @@ internal class TradeInputCalculator(
                 "openPositions.$marketId",
             ),
         )
-        return maxLeverageFromPosition(position, market)
+        val equity = parser.asDouble(parser.value(subaccount, "equity.current"))
```
mike-dydx, line 1691: nit: can do the `?: return null` here instead of checking later

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
```diff
@@ -1300,7 +1275,6 @@ internal class TradeInputCalculator(
                 account,
                 subaccount,
                 trade,
-                position,
                 market,
```
mike-dydx, line 1278: how come `calculatedOptions` does not need `subaccount` as an arg now?`

moo-onthelawn: do u mean `position`? If you trace back [here](https://github.com/dydxprotocol/v4-abacus/pull/585/files/4d94186a37f526ea449f53965ffa14a765a024ab#diff-25886307b00a8d85e0a52116d3569ca56fb322bee615ed7731009848b25361c7R168) it's cuz the `maxLeverage` function here (which replaced `maxLeverageFromPosition`) just reads the position from the passed through `subaccount` when it needs it (instead of earlier on). And the reason I passed through subaccount is cuz we now need the equity and freeCollateral off of subaccount as well

mike-dydx (approved):

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
mike-dydx: oh, i missed `subaccount` is already an earlier argument. Nevermind!

ruixhuang (approved):

PR #586, opened 2024-08-21
Enable svm on mainnet

tian000:
This PR enables support for SVM chains on mainnet


yogurtandjam (dismissed):

yogurtandjam:
[hold on merging] we need to filter out non solana chains when solana is selected as the wallet type.

tian000:
@yogurtandjam here's a PR that prevents solana from showing up for EVM wallets: https://github.com/dydxprotocol/v4-web/pull/965/files

rosepuppy:
We have a way of filtering for ChainType in abacus see:
https://github.com/dydxprotocol/v4-abacus/blob/251589093b8bdcf472d84a83a3289843afd390ec/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipProcessor.kt#L32

But this requires some prework of letting abacus know that Solana is connected.

I recognize that https://github.com/dydxprotocol/v4-web/pull/965 should filter out the appropriate chain source on web, but ideally we can we do the filtering in abacus, so that logic is shared amongst all clients.

rosepuppy:
Ideally we have SVM filtering in abacus, but I'll approve for now to unblock, but will add a follow up task to make sure it is tracked.

rosepuppy (approved):

PR #588, opened 2024-08-21
feat: call compliance endpoint for keplr

rosepuppy:
Backend endpoint PR: https://github.com/dydxprotocol/v4-chain/pull/2117

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
@@ -751,7 +751,7 @@ internal open class AccountSupervisor(
         complianceAction: ComplianceAction
     ) {
         val chainId = helper.environment.dydxChainId
-        val message = "Compliance verification message"
+        val message = "Verify account ownership"
         val payload =
             helper.jsonEncoder.encode(
                 mapOf(
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
@@ -778,7 +778,9 @@ internal open class AccountSupervisor(
                 val isUrlAndKeysPresent =
                     url != null &&
                         signedMessage != null &&
-                        publicKey != null && (timestamp != null || isKeplr == true)
+                        publicKey != null
+
+                val isKeplrOrHasTimestamp = (timestamp != null || isKeplr == true)
 
                 val isStatusValid = status != ComplianceStatus.UNKNOWN
 
@@ -803,7 +805,7 @@ internal open class AccountSupervisor(
                         )
                     }
 
-                if (isUrlAndKeysPresent && isStatusValid) {
+                if (isUrlAndKeysPresent && isKeplrOrHasTimestamp && isStatusValid) {
                     val header =
                         iMapOf(
                             "Content-Type" to "application/json",
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
@@ -784,28 +784,27 @@ internal open class AccountSupervisor(
 
                 val isStatusValid = status != ComplianceStatus.UNKNOWN
 
-                val body: Map<String, String> =
-                    if (isKeplr == true) {
-                        iMapOf(
-                            "address" to address.rawAddress,
-                            "message" to message,
-                            "action" to complianceAction.toString(),
-                            "signedMessage" to signedMessage!!,
-                            "pubkey" to publicKey!!,
-                        )
-                    } else {
-                        iMapOf(
-                            "address" to address.rawAddress,
-                            "message" to message,
-                            "currentStatus" to status.toString(),
-                            "action" to complianceAction.toString(),
-                            "signedMessage" to signedMessage!!,
-                            "pubkey" to publicKey!!,
-                            "timestamp" to timestamp!!,
-                        )
-                    }
-
                 if (isUrlAndKeysPresent && isKeplrOrHasTimestamp && isStatusValid) {
+                    val body: Map<String, String> =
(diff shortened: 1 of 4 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -767,20 +769,29 @@ internal open class AccountSupervisor(
             val result = helper.parser.decodeJsonObject(additionalPayload)
 
             if (error == null && result != null) {
-                val url = complianceGeoblockUrl()
                 val signedMessage = helper.parser.asString(result["signedMessage"])
                 val publicKey = helper.parser.asString(result["publicKey"])
                 val timestamp = helper.parser.asString(result["timestamp"])
+                val isKeplr = helper.parser.asBool(result["isKeplr"])
+                val url = if (isKeplr == true) complianceGeoblockKeplrUrl() else complianceGeoblockUrl()
```
yogurtandjam, line 776: [discussion] we're starting to get a lot of branching in this function. you think it's worth splitting this into 2 separate functions? or creating 2 separate helper functions?

if this is urgent i'm down to stamp right away though since the logic makes sense

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -767,20 +769,29 @@ internal open class AccountSupervisor(
...
 
                 val isUrlAndKeysPresent =
                     url != null &&
                         signedMessage != null &&
-                        publicKey != null &&
-                        timestamp != null
+                        publicKey != null && (timestamp != null || isKeplr == true)
```
yogurtandjam, line 781: [fix/suggestion] would prefer to separate the `timestamp!=null || isKelper == true` into a separately named variable, or rename `isUrlAndKeysPresent` to explain more of what's happening here.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -767,20 +769,29 @@ internal open class AccountSupervisor(
...
 
-                if (isUrlAndKeysPresent && isStatusValid) {
-                    val body: IMap<String, String> =
+                val body: Map<String, String> =
+                    if (isKeplr == true) {
+                        iMapOf(
+                            "address" to address.rawAddress,
+                            "message" to message,
+                            "action" to complianceAction.toString(),
+                            "signedMessage" to signedMessage!!,
```
yogurtandjam, line 791: [q] why do these properties have to be assumed non-null? (i know it's part of previously committed code too, but curious if you have the answer)

rosepuppy: Just that the request needs all these fields.

I check for it to be not null with `isUrlAndKeysPresent` before the body gets used, so if anything is null, the request would not send.

I don't know if there is a better kotlin way to do this tho

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
rosepuppy: Definitely thought about it, and tried initially. Ended with this approach because the only code path change is just the body and url, but if you feel strongly I can take another stab!

yogurtandjam: nah no worries! i think we can do this as part of a larger cleanup task in the future to reduce code duplication. the `OnboardingSupervisor` is a really bad offender of this... so easy to make mistakes when updating things since there's so much duplicate code

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
yogurtandjam: oh, does kotlin remove the field entirely if the value is null?

rosepuppy: oh it will actually throw, thats a good catch actually. I think it should be inside the if statement

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
rosepuppy: yea, feel that

yogurtandjam (approved):
stamped but i think we need to wait for https://github.com/dydxprotocol/v4-chain/pull/2117/files to merge right? so just ping me when you need a restamp (you will, most likely, as you'll have to bump versions)

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
```diff
@@ -784,28 +784,27 @@ internal open class AccountSupervisor(
...
-                    }
-
                 if (isUrlAndKeysPresent && isKeplrOrHasTimestamp && isStatusValid) {
+                    val body: Map<String, String> =
+                        if (isKeplr == true) {
+                            iMapOf(
+                                "address" to address.rawAddress,
+                                "message" to message,
+                                "action" to complianceAction.toString(),
+                                "signedMessage" to signedMessage!!,
```
yogurtandjam, line 794: [nit] we shouldn't need the non-null casting anymore, right?

rosepuppy:
> stamped but i think we need to wait for https://github.com/dydxprotocol/v4-chain/pull/2117/files to merge right? so just ping me when you need a restamp (you will, most likely, as you'll have to bump versions)

Should be fine, since the request won't get called unless the user is using Keplr which isn't launched yet

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/AccountSupervisor.kt
rosepuppy: still need to do for the type check haha

PR #589, opened 2024-08-21
feat: limit close orders with default mid-market price as limit price

aforaleka:
gated by statsig flag `ff_enable_limit_close`, this is to enable closing position via limit order
UI is tbd, but what's needed on web/mobile is an option to switch between Market / Limit order, and an input to specify limit price. this shouldn't be enabled until protocol adds support

- add `useLimit` and `limitPrice` to close position input
- add mid market price calculation helper fn. when `useLimit` is set to true, set to limit price to mid market price by default. (toggling this option just reset it to mid market price by design)
- update close position payload to support limit order type (default to GTT 28 days when limit)

testing:
- updated close position unit tests
- tested locally for v1
- todo: v2 static typing isn't working for close position inputs already t_t will get another PR for that
image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -52,7 +52,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
-version = "1.8.99"
+version = "1.8.100"
 
 repositories {
     google()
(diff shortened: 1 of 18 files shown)


ruixhuang (dismissed):

ruixhuang:
I will be adding static typing for close position shortly, so it's nice to have this now!

prashanDYDX (approved):

PR #592, opened 2024-08-22
fix showing one sided liquidity error when it should not

aforaleka:
it depends on summary.slippage. but summary can be a non-null empty object.

image

image

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/validator/trade/TradeMarketOrderInputValidator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/validator/trade/TradeMarketOrderInputValidator.kt
@@ -141,7 +141,8 @@ internal class TradeMarketOrderInputValidator(
             )
         }
 
-        val summary = parser.asNativeMap(trade["summary"]) ?: return null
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/validator/trade/TradeMarketOrderInputValidator.kt
```diff
@@ -141,7 +141,8 @@ internal class TradeMarketOrderInputValidator(
             )
         }
 
-        val summary = parser.asNativeMap(trade["summary"]) ?: return null
+        // summary can be empty even though there's input e.g. leverage input is set to the same as current position leverage
+        val summary = parser.asNativeMap(trade["summary"])?.takeIf { it.isNotEmpty() } ?: return null
```
aforaleka, line 145: i hope the typed version avoids this t_t? but it's also where summary is set

ruixhuang: It should be fine.  The typed version has `trade.summary` being either null or filled with data.

ruixhuang (approved):

ruixhuang (approved):

PR #594, opened 2024-08-22
feat: vault form validation

tyleroooo:


src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultDepositWithdrawForm.kt
```diff
@@ -0,0 +1,274 @@
+package exchange.dydx.abacus.functional.vault
+
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+
+@JsExport
+@Serializable
+data class FormData(
```
ruixhuang, line 8: Can we use prefix those data classes with "Vault" (i.e., VaultFormData, VaultTransactionAction, etc)?  This helps minimizing name-conflicts.

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultDepositWithdrawForm.kt
```diff
@@ -0,0 +1,274 @@
...
+@JsExport
+@Serializable
+data class SlippageResponse(
+    val shares: Double,
+    val expectedAmount: Double,
+)
+
+@JsExport
+@Serializable
+data class ValidationError(
```
ruixhuang, line 38: Can we reuse `ValidationError` in `exchange.dydx.abacus.output.input` ?  It has the fields to support localization of error messages.

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultDepositWithdrawForm.kt
tyleroooo: oh yeah that's a good idea. Unfortunate that they're all globally scoped.

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultDepositWithdrawForm.kt
tyleroooo: Hmmm I kind of like this smaller type for this case. It's nice that the errors are an enum so the frontends can guarantee they are handling all of the possibilities when abacus upgrades. Most of the fields on the other error object are useless for us here too.

tyleroooo: In general I don't really like abacus knowing anything about localization

jaredvu: For mobile clients, their current pattern is to receive fully localized strings to render directly. @ruixhuang @mike-dydx is it possible to handle the enums within client code to map to strings?

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultDepositWithdrawForm.kt
```diff
@@ -0,0 +1,274 @@
...
+    val severity: VaultFormValidationErrorSeverity,
+    val type: VaultFormValidationErrorType
+)
+
+@JsExport
+@Serializable
+enum class VaultFormValidationErrorSeverity {
+    WARNING,
+    ERROR
+}
```
jaredvu, line 48: I don't think we need to add ErrorSeverity for every form that we add. There is an existing `ErrorType` from `output/input/ValidationError` that includes error, warning, required.

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultDepositWithdrawForm.kt
prashanDYDX: unrelated to localization - we could also use a `sealed` type (think union type in TS) for this:

```
sealed interface ValidationResult

data class Success(
  val submissionData 
  ...
) : ValidationResult

data class Error(
  val errors: ...
) : ValidationResult
```


on the localization topic, yeah it's ofc doable, just more boilerplate to write and a bit more work to keep in sync if things change later.

tyleroooo: I don't think it's a great fit here since one of the errors is a warning (the one about slippage) so present even when things are fine.

ruixhuang (approved):

jaredvu (approved):

PR #597, opened 2024-08-23
TRCL-3644 Static Typing: TriggerOrder/SL/TP

ruixhuang:
* Updating trigger order related classes
* Fixing an issue of the ClosePosition order payload with the static typing code path.
* Tested on Android and guarded by feature flag.


moo-onthelawn (dismissed):
🔥

prashanDYDX (approved):

PR #598, opened 2024-08-23
fix(isolatedMargin): update percentages on margin adjustment to be relative to total margin instead of existing position

moo-onthelawn:
## Context
Adjusting margin should be a % of maximum amount of margin you can take out, not a % of your position’s margin. (i.e. 100% means remove 100% of margin that you are able to take out. This puts you at that market’s max leverage (or close to))

oops branch name should actually be `mulan/ct-1109-abacus-update-margin-calculations`

## FE proof of concept 
Notes:
- Ideally I get validation into abacus instead of having it separate on FE cuz half these errors are just v confusing
- due to constantly changing vals (and maybe rounding errors), sometimes 100% causes an error - can dig in a lil more before pushing the FE change
  - not a big deal cuz current FE only has 75% option

https://github.com/user-attachments/assets/108e5bda-effa-44dc-9c4e-3cb4d9164fea

### AdjustIsolatedMarginInput
- added `marketId` to input object 
  - necessary to calculate the market's max leverage (and how much margin we can remove)
- added `amountInput` to input object
  - represents whether the user last input via the percentage button or an amount. Determines how we calculate both fields
  - in an ideal world we'd maybe have an `amount: Amount` object where the `Amount` object has `amount`, `amountPercent`, and `amountInput` but this would be a breaking change and I don't forsee this getting more complex

### TradingStateMachine+AdjustIsolatedMarginInput
- Move logic of setting percent/usdc amount into `AdjustIsolatedMarginInputCalculator`
- Add `Market` to `AdjustIsolatedMarginInputField` so it can be set by caller (FE)
  - Add logic to `TradingStateMachine.adjustIsolatedMargin` for setting `market`
- Update the setting of `AmountPercent` and `Amount` to also set `AmountInput` so that the calculator knows whether to calculate percentage off of Amount (or vice versa)

### AdjustIsolatedMarginInputCalculator
- Add `calculateAmounts` helper which sets percent/usdc amount based on last amount input

## Documentation Updates
- `docs/Input.md`
- `docs/Input/AdjustIsolatedMarginInput.md`

TODO:
- [ ] Mobile instructions (this is a breaking change)
  - [ ] When pulling in this update, mobile will **need** to set the market from their FE when this dialog pops up, otherwise the calculations will not run (unsure how to make this breaking / it'd be nice if we could auto-set huh) [example](https://github.com/dydxprotocol/v4-web/pull/947/files#diff-af6704347ab5a1b9f6a07768c6f04896ef5b8e82e97a4d22fef1311626a4a48aR80)
- [x] update documentation
- [ ] ~~apply to both versions~~ (there's no second version)
- [x] write tests


aforaleka (approved):
thx for updating the docs!

PR #602, opened 2024-08-26
TRCL-3636 Static Typing: Adjust Isolated Margin Input

ruixhuang:
Updating Adjust Isolated Margin Input related classes
Adding fixing from [PR](https://github.com/dydxprotocol/v4-abacus/pull/598).


src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
```diff
@@ -129,6 +130,26 @@ data class AdjustIsolatedMarginInput(
...
+        internal fun create(
+            parser: ParserProtocol,
+            data: InternalAdjustIsolatedMarginInputState?
+        ): AdjustIsolatedMarginInput? {
+            return if (data != null) {
+                AdjustIsolatedMarginInput(
+                    market = data.market,
+                    type = data.type ?: IsolatedMarginAdjustmentType.Add,
+                    amount = parser.asString(data.amount),
+                    amountPercent = parser.asString(data.amountPercent),
```
moo-onthelawn, line 142: maybe a dumb question but why do we still need `parser` here and the line below?

src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/AdjustIsolatedMarginInputCalculatorV2.kt
```diff
@@ -0,0 +1,207 @@
...
+            crossFreeCollateral = parentSubaccount?.calculated?.get(CalculationPeriod.current)?.freeCollateral,
+            crossFreeCollateralUpdated = parentSubaccount?.calculated?.get(CalculationPeriod.post)?.freeCollateral,
+            crossMarginUsage = parentSubaccount?.calculated?.get(CalculationPeriod.current)?.marginUsage,
+            crossMarginUsageUpdated = parentSubaccount?.calculated?.get(CalculationPeriod.post)?.marginUsage,
+            positionMargin = position?.calculated?.get(CalculationPeriod.current)?.marginValue,
+            positionMarginUpdated = position?.calculated?.get(CalculationPeriod.post)?.marginValue,
+            positionLeverage = position?.calculated?.get(CalculationPeriod.current)?.leverage,
+            positionLeverageUpdated = position?.calculated?.get(CalculationPeriod.post)?.leverage,
+            liquidationPrice = position?.calculated?.get(CalculationPeriod.current)?.liquidationPrice,
+            liquidationPriceUpdated = position?.calculated?.get(CalculationPeriod.post)?.liquidationPrice,
```
moo-onthelawn, line 204: super nit, I wonder if this would be easier to read if we had `postPosition` and `currentPosition` variables

moo-onthelawn (approved):

src/commonMain/kotlin/exchange.dydx.abacus/output/input/AdjustIsolatedMarginInput.kt
ruixhuang: Those fields are defined as string type in `AdjustIsolatedMarginInput`, which is frontend-facing.  It would be a breaking change if we change the data type.

src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/AdjustIsolatedMarginInputCalculatorV2.kt
ruixhuang: Can update in the future.... and then there are two calculated fields for subaccount and position, so it gets pretty confusing.

PR #610, opened 2024-08-29
TRCL-3643 Static Typing: TransferInput

ruixhuang:
- Updating Transfer Input for static typing.  The Squid/Skip payload is still parsed as map, and can be converted in the future if we have time.
- Updating Withdrawal gating for static typing.
- Updating all remaining tests.  You can now set staticTyping = true in BaseTests to test the new codepath.
- Tested on Android


src/commonMain/kotlin/exchange.dydx.abacus/processor/input/TransferInputProcessor.kt
```diff
@@ -0,0 +1,371 @@
...
+            .copy(assets = tokenOptions.toIList())
+        transfer.withdrawalOptions = WithdrawalInputOptions.safeCreate(transfer.withdrawalOptions)
+            .copy(assets = tokenOptions.toIList())
+    }
+
+    private fun updateTransferToTokenType(
+        transfer: InternalTransferInputState,
+        tokenAddress: String
+    ) {
+        // Code not used
```
rosepuppy, line 330: Can you expand on when this code is not being used? 🤔

ruixhuang: Line 333 to 355 is already called before this function is called.

`resources.tokenSymbol `and `resources.tokenDecimals` are not read, so there is no need to set them.

PR #611, opened 2024-08-29
add cancel orders with tx queue

aforaleka:
add functionality to cancel all orders (either in all markets or a single market)

(originally wanted to use batch cancel for short term orders, but integration was a bit more complicated (due to needing tx broadcast mode to get success/failed order ids, and canceling short term orders via FE almost always fails). therefore will reconsider when batch cancel also support stateful orders.

-> decided to just rely on TX queue to submit cancels one by one. short term cancels don't need txcommit so they will go first. we could potentially run into rate limiting issue here but decided that it'll be rare for retail users.)

testing:
- unit test
- tested locally on web

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
@@ -14,6 +14,7 @@ import exchange.dydx.abacus.responses.ParsingErrorType
 import exchange.dydx.abacus.responses.ParsingException
 import exchange.dydx.abacus.state.manager.CancelOrderRecord
 import exchange.dydx.abacus.state.manager.FaucetRecord
+import exchange.dydx.abacus.state.manager.HumanReadableCancelMultipleOrdersPayload
 import exchange.dydx.abacus.state.manager.HumanReadableCancelOrderPayload
 import exchange.dydx.abacus.state.manager.HumanReadablePlaceOrderPayload
 import exchange.dydx.abacus.state.manager.HumanReadableSubaccountTransferPayload
@@ -31,6 +32,7 @@ import exchange.dydx.abacus.utils.IMap
 import exchange.dydx.abacus.utils.IMutableList
 import exchange.dydx.abacus.utils.Logger
 import exchange.dydx.abacus.utils.NUM_PARENT_SUBACCOUNTS
+import exchange.dydx.abacus.utils.Numeric
 import exchange.dydx.abacus.utils.ParsingHelper
 import exchange.dydx.abacus.utils.SHORT_TERM_ORDER_FLAGS
 import exchange.dydx.abacus.utils.iMapOf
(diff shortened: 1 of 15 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
```diff
@@ -147,6 +148,36 @@ internal class SubaccountTransactionSupervisor(
         return submitCancelOrder(orderId, marketId, callback, payload, analyticsPayload, uiClickTimeMs)
     }
 
+    internal fun cancelAllOrders(marketId: String?, callback: TransactionCallback): HumanReadableCancelAllOrdersPayload {
+        val payload = payloadProvider.cancelAllOrdersPayload(marketId)
+
+        payload.payloads.forEach { cancelPayload ->
+            val subaccount = stateMachine.state?.subaccount(subaccountNumber)
```
moo-onthelawn, line 155: nit: `val subaccount = stateMachine.state?.subaccount(subaccountNumber) ?: return null` maybe? since subaccount should never be null

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
```diff
@@ -147,6 +148,36 @@ internal class SubaccountTransactionSupervisor(
...
+            val existingOrder = subaccount?.orders?.firstOrNull { it.id == cancelPayload.orderId }
+                ?: throw ParsingException(
+                    ParsingErrorType.MissingRequiredData,
+                    "no existing order to be cancelled for $cancelPayload.orderId",
```
moo-onthelawn, line 159: hmm is it possible to hit this via race condition (i.e. if a user has two tabs open, clicks "cancel all orders" in one tab, and then manually cancels an existing order in another)? Instead of throwing an error, is it possible to just log it and gracefully continue?

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
```diff
@@ -147,6 +148,36 @@ internal class SubaccountTransactionSupervisor(
...
+                )
+            val analyticsPayload = analyticsUtils.cancelOrderAnalyticsPayload(
```
moo-onthelawn, line 161: nit: can we use named args here

moo-onthelawn (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
ruixhuang: Let Android Studio do this for you:

https://github.com/user-attachments/assets/553d8728-a2d2-4e5f-8588-45b3ae6c2798

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
aforaleka: order should still exists even if it's canceled!

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
aforaleka: that's why the exception is thrown!

aforaleka: but actually u right i can move this up!!!

ruixhuang (approved):

PR #613, opened 2024-08-30
chore: update default target leverage on isolated market orders to be max market leverage

moo-onthelawn:
instead of defaulting to 1 as target leverage, default to max market lev

still respects existing position leverage, if it exists

https://github.com/user-attachments/assets/07337cca-d4ee-4ed6-8a80-c6de97a3969a


src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
```diff
@@ -78,7 +78,6 @@ internal class TradeInputCalculator(
         val subaccount = if (marginMode == MarginMode.Cross) {
             crossMarginSubaccount
         } else {
-            // TODO: incorrect for isolated trades; fix CT-1092
```
moo-onthelawn, line 81: not actually incorrect, sorry poorly placed TODO which was just to remind me to properly calculate maxLeverage for isolated which we no longer need

src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarginModeCalculator.kt
```diff
@@ -19,6 +20,16 @@ internal class TradeInputMarginModeCalculator {
             marketId = tradeInput.marketId,
             subaccountNumber = subaccountNumber,
         )
+        val market = markets?.get(tradeInput.marketId)
+        val imf = market?.perpetualMarket?.configs?.initialMarginFraction ?: Numeric.double.ZERO
```
ruixhuang, line 24: Nit: Since this logic is used elsewhere, we can make it a computed property of `MarketConfigs`... Something like this:

```
data class MarketConfigs(...) {
 ....
 
     internal val maxMarketLeverage: Double
        get() {
            val imf = initialMarginFraction ?: Numeric.double.ZERO
            val effectiveImf = effectiveInitialMarginFraction ?: Numeric.double.ZERO
            return if (effectiveImf > Numeric.double.ZERO) {
                Numeric.double.ONE / effectiveImf
            } else if (imf > Numeric.double.ZERO) {
                Numeric.double.ONE / imf
            } else {
                Numeric.double.ONE
            }
        }
}
```

ruixhuang (approved):

src/commonMain/kotlin/exchange.dydx.abacus/calculator/MarginModeCalculator.kt
```diff
@@ -15,6 +16,18 @@ object MarginModeCalculator {
...
+
+        val imf = parser.asDouble(parser.value(market, "configs.initialMarginFraction")) ?: Numeric.double.ZERO
+        val effectiveImf = parser.asDouble(parser.value(market, "configs.effectiveInitialMarginFraction")) ?: Numeric.double.ZERO
+        val maxMarketLeverage = if (effectiveImf > Numeric.double.ZERO) {
+            Numeric.double.ONE / effectiveImf
+        } else if (imf > Numeric.double.ZERO) {
+            Numeric.double.ONE / imf
+        } else {
+            Numeric.double.ONE
+        }
```
jaredvu, line 29: the calculation for `maxMarketLeverage` is duplicated in quite a few places. Would it be easier to have a pure function in `marketCalculator` or something similar and just use in each of these cases?

jaredvu: Similar to what Rui said, if it can be included in the actual configs struct it would be easy to de-dupe this calculation

moo-onthelawn:
closed due to unsigned commit from mobile bot https://github.com/dydxprotocol/v4-abacus/pull/631

PR #624, opened 2024-09-05
Set up Publish action in Github Actions

prashanDYDX:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          distribution: 'adopt'
+          java-version: '17'
+      - uses: gradle/actions/setup-gradle@v4
+
+      - name: Publish to NPM
+        run: |
+          ./gradlew assembleJsPackage
+          cp "LICENSE" "build/packages/js/LICENSE"
+          ./gradlew publishJsPackageToNpmjsRegistry
```
jaredvu, line 77: Should we re-add the `./gradlew clean` and `./gradlew packJsPackage` commands?

prashanDYDX: they're not necessary - we haven't done a build so clean won't do anything. and `publishJsPackageToNpmjsRegistry` will inherently call `packJsPackage`

jaredvu (approved):

PR #625, opened 2024-09-05
Expose current PerpetualState directly to clients.

prashanDYDX:


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
```diff
@@ -76,8 +77,8 @@ class AsyncAbacusStateManagerV2(
         }
     }
 
-    private val environmentsFile = ConfigFile.ENV
-    private val documentationFile = ConfigFile.DOCUMENTATION
```
jaredvu, line 80: Were these deletions on purpose? Unused?

jaredvu (dismissed):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
prashanDYDX: yah unused

ruixhuang (dismissed):

PR #628, opened 2024-09-08
TRCL-3643 Static Typing: TransferInput

ruixhuang:
- Updating Transfer Input for static typing. The Squid/Skip payload is still parsed as map, and can be converted in the future if we have time.
- Updating Withdrawal gating for static typing.
- Updating all remaining tests. You can now set staticTyping = true in BaseTests to test the new codepath.
- Tested on Android


src/commonMain/kotlin/exchange.dydx.abacus/processor/input/TransferInputProcessor.kt
```diff
@@ -0,0 +1,371 @@
...
+            inputState.transfer.address = null
+            inputState.transfer.memo = null
+            inputState.transfer.depositOptions = null
+            inputState.transfer.withdrawalOptions = null
+            inputState.transfer.transferOutOptions = null
+            inputState.transfer.summary = null
+            inputState.transfer.resources = null
+            inputState.transfer.route = null
+
+            inputState.transfer.type = TransferType.deposit
```
jaredvu, line 64: is this transferType just the `default` before it gets overwritten in the `when` block?

ruixhuang: Yes, it's overwritten only when the inputField is "TransferInputField.type".   It's the same logic as before.

rosepuppy (approved):

src/commonMain/kotlin/exchange.dydx.abacus/processor/input/TransferInputProcessor.kt
jaredvu: Sounds good! Didn't realize that we defaulted to a transferType instead of leaving as null

jaredvu (approved):

PR #633, opened 2024-09-09
TRCL-3682 Add isNew to PerpetualMarket

ruixhuang:
Fetching 7 day sparkline to determine if the market is new for mobile apps.  Web already does this outside Abacus, and it's currently configured to fetch for mobile apps only.  But we can remove the web code and use the isNew field from Abacus in the future.

Also updating the 1 day sparkline fetch interval from 1 minute to 1 hour per Slack discussion.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/Configs.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/Configs.kt
@@ -43,6 +43,7 @@ data class MarketsConfigs(
     val subscribeToOrderbook: Boolean,
     val subscribeToTrades: Boolean,
     val subscribeToCandles: Boolean,
+    val retrieveSevenDaySparkline: Boolean,
 ) {
     companion object {
         val forApp = MarketsConfigs(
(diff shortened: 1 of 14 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/Configs.kt
```diff
@@ -62,6 +64,7 @@ data class MarketsConfigs(
             subscribeToOrderbook = true,
             subscribeToTrades = true,
             subscribeToCandles = false,
+            retrieveSevenDaySparkline = false,
```
aforaleka, line 67: should we set this to true so we can assume web will use the isNew field?

aforaleka (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/Configs.kt
ruixhuang: If it's set to `true`, the Abacus would do the fetch.  It's better to turn it `true` after the web fetch is turned off, so that we don't fetch the same thing twice.

PR #638, opened 2024-09-10
add new balancePercent input to trade size inputs

moo-onthelawn:
Adds a `balancePercent` to trade size inputs such that user can choose an amount for their trade based on the percentage of their available balance (we're replacing the 1x...10x leverage buttons with this percentage toggle). See screen recording proof of concept:

https://github.com/user-attachments/assets/c8562fa9-508d-435c-bd59-53cc64398f9b

For now, this only applies to cross market orders. Will follow up in separate PR to enable for isolated market orders (will probably choose to use a `needsPercentBalance` field). Will likely not enable for limit orders until leverage slider is added. 

Note this change is non breaking: to get this, caller just has to update to call into this new field (and set appropriately).

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarketOrderCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarketOrderCalculator.kt
@@ -106,13 +106,7 @@ internal class TradeInputMarketOrderCalculator() {
         val tradeSize = trade.size
         if (tradeSize != null) {
             val freeCollateral = subaccount?.calculated?.get(CalculationPeriod.current)?.freeCollateral ?: return null
-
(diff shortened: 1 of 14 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarketOrderCalculator.kt
```diff
@@ -89,11 +105,16 @@ internal class TradeInputMarketOrderCalculator() {
     ): TradeInputMarketOrder? {
         val tradeSize = trade.size
         if (tradeSize != null) {
+            val freeCollateral = subaccount?.calculated?.get(CalculationPeriod.current)?.freeCollateral ?: return null
```
aforaleka, line 108: do we need to add a check for free collateral > 0 as well?

aforaleka (dismissed):
just the nit about potentially safer checks for not division by zero!

ruixhuang (approved):

PR #642, opened 2024-09-10
Static Typing: Clean up market processor codepath and update tests

ruixhuang:
- Removing duplicate codepath for market data.
- Updated existing tests.
- Tested on Android.


src/commonMain/kotlin/exchange.dydx.abacus/state/internalstate/InternalState.kt
```diff
@@ -173,6 +173,10 @@ internal data class InternalTradeInputOptions(
 
 internal data class InternalMarketSummaryState(
     var markets: MutableMap<String, InternalMarketState> = mutableMapOf(),
+
+    var volume24HUSDC: Double? = null,
+    var openInterestUSDC: Double? = null,
+    var trades24H: Double? = null,
```
prashanDYDX, line 179: these seem like they could default to 0 and be non-nullable

ruixhuang: There is a difference though.  null means the value is pending

PR #650, opened 2024-09-13
fix balancePercent calculations for trades on opposite side of current position

moo-onthelawn:
fixed a couple bugs with balancePercent calculations. math will be the end of me I swear
- Bug 1: Fix balancePercent size calculation for cross market trades on the opposite side of current position
  - The balance percent a user selects should include closing their whole current position and then apply the percentage off of their new free collateral given the entire closed position
- Bug 2: Fix balancePercent size calculation for isolated market trades on opposite side of current position
  - It doesn't make sense to apply balancePercent on the freeCollateral since it doesn't impact their free collateral - make balance percent apply to the currentPosition size instead
- Bug 3: Fix balancePercent seemingly being off for isolated market trades on same side of current position
  - It's cuz isolated markets need to do an additional calculation for the amount of additional collateral to trade when adding to current position size
  - Note: in ideal world I'd be using shared functions in MarginCalculator but I need to do a bit of refactoring to get the correct subaccount/childsubaccount variables. Filed a TODO to fix it.
  - Was able to adjust the collateral buffer percents after that fix

proof of concept help:

https://github.com/user-attachments/assets/7ec0ef78-05f4-4423-ac22-bd036f149612


src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarketOrderCalculator.kt
```diff
@@ -569,6 +633,24 @@ internal class TradeInputMarketOrderCalculator() {
...
+    private fun calculateBalancePercentFromUsdcSize(
+        usdcSize: Double,
+        freeCollateral: Double,
+        positionSize: Double,
+        tradeLeverage: Double,
+        isTradeSameSide: Boolean,
+    ): Double {
+        if (freeCollateral <= Numeric.double.ZERO) {
+            return Numeric.double.ZERO
+        }
```
jaredvu, line 645: would it make sense to add a non-zero check for `tradeLeverage` as well?

src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarketOrderCalculator.kt
```diff
@@ -558,7 +622,7 @@ internal class TradeInputMarketOrderCalculator() {
                 break@orderbookLoop
             }
         }
-        val balancePercentTotal = (usdcSizeTotal / tradeLeverage) / freeCollateral
+        val balancePercentTotal = calculateBalancePercentFromUsdcSize(usdcSize = usdcSizeTotal, freeCollateral = freeCollateral, positionSize = positionSizeNotional, tradeLeverage = tradeLeverage, isTradeSameSide = isTradeSameSide)
```
jaredvu, line 625: im surprised `spotless` doesn't format this to multiple lines

src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarketOrderCalculator.kt
```diff
@@ -118,11 +119,28 @@ internal class TradeInputMarketOrderCalculator() {
                 maxMarketLeverage
             }
 
+            val positions = subaccount.openPositions
+            val marketId = market?.perpetualMarket?.id
```
jaredvu, line 123: it looks like you are getting `marketId` from `trade.marketId` in the non-typed `TradeInputCalculator`. I'm assuming its interchangeable

moo-onthelawn: ah yeah - they should be interchangeable, but lemme make them consistent for easier to read reasons

jaredvu (approved):

PR #652, opened 2024-09-15
Add solanaWalletConnected

tian000:
Adds solanaWalletConnected the same way that cosmosWalletConnected is implemented

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/AsyncAbacusStateManagerProtocol.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/AsyncAbacusStateManagerProtocol.kt
@@ -111,6 +111,7 @@ interface AsyncAbacusStateManagerSingletonProtocol {
     var subaccountNumber: Int
     var market: String?
     var cosmosWalletConnected: Boolean?
+    var solanaWalletConnected: Boolean?
 }
 
 @JsExport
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/AsyncAbacusStateManagerProtocol.kt
```diff
@@ -111,6 +111,7 @@ interface AsyncAbacusStateManagerSingletonProtocol {
     var subaccountNumber: Int
     var market: String?
     var cosmosWalletConnected: Boolean?
+    var solanaWalletConnected: Boolean?
```
ruixhuang, line 114: Can this be refactored as an enum, since the wallet connection types are mutually exclusive?

```
enum class WalletConnectionType {
    Ethereum,
    Cosmos,
    Solana
}

...
var walletConnectionType: WalletConnectionType?
```

tian000:
Closing in favor of: https://github.com/dydxprotocol/v4-abacus/pull/656

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/AsyncAbacusStateManagerProtocol.kt
tian000: Done: https://github.com/dydxprotocol/v4-abacus/pull/656

PR #657, opened 2024-09-16
TRCL-3695 Add configs for notification provider types

ruixhuang:
By default, all provider types are enabled, but the client can optionally set the notification types it wants.

This is needed for mobile push notifications where the fills and order status notifications will come from backend instead of Abacus.


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/notification/NotificationsProvider.kt
```diff
@@ -6,51 +6,77 @@ import exchange.dydx.abacus.state.manager.V4Environment
...
-            useParentSubaccount,
-        ),
-        OrderStatusChangesNotificationProvider(
-            stateMachine,
-            uiImplementations,
-            parser,
-            jsonEncoder,
-        ),
+    val notifications: List<NotificationProviderType> = listOf(
+        NotificationProviderType.BlackReward,
```
mike-dydx, line 30: ⚫ Blockreward ⚫

src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/DynamicLocalizer.kt
```diff
@@ -251,8 +251,7 @@ class DynamicLocalizer(
                     }
                 }
             } else {
-                // Not a valid language
-                callback(false, null)
+                callback(_loadingLanguage == language, null)
```
ruixhuang, line 254: This fixes an existing race condition of language settings on iOS.

mike-dydx (approved):

PR #660, opened 2024-09-17
feat: market close all positions

aforaleka:
adding a helper function to market close all open positions under parent subaccount 0

in the future when indexer supports mid price ws endpoint, and protocol support limit close (reduce-only non-IOC), we'll only add a limit close option. until then, this is just full closing all positions with market price (i.e. oracle price + slippage)

tested locally https://github.com/dydxprotocol/v4-web/pull/1031

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionPayloadProvider.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionPayloadProvider.kt
@@ -1,17 +1,22 @@
 package exchange.dydx.abacus.state.v2.supervisor
 
 import abs
+import exchange.dydx.abacus.calculator.CalculationPeriod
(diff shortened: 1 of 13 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionPayloadProvider.kt
```diff
@@ -268,6 +286,19 @@ internal class SubaccountTransactionPayloadProvider(
         return HumanReadableTriggerOrdersPayload(marketId, positionSize, placeOrderPayloads, cancelOrderPayloads)
     }
 
+    private fun subaccountNumberForPosition(marketId: String): Int {
```
prashanDYDX, line 289: don't we have another helper for this somewhere already? @jaredvu

ruixhuang (approved):

ruixhuang (approved):

PR #664, opened 2024-09-19
fix: fix isolated pnl transfer calc

moo-onthelawn:
noticed a bug when testing this feature - isolated margin percent inputs looked wrong esp when the amount was high - realized it was cuz the collateral transfer calculation was wrong - we need to include the isolatedPnlImpact in the balance used

(for some reason when I build locally now it keeps telling the testnet chain is halted? I tried with building a local abacus with no change but an extra space and it still said halted so assuming this is not related)

https://github.com/user-attachments/assets/5323f9e2-7ea5-4347-aa78-045ab320cc28

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
@@ -25,6 +25,7 @@ import exchange.dydx.abacus.utils.safeSet
 import kollections.JsExport
 import kotlinx.serialization.Serializable
 import kotlin.math.abs
+import kotlin.math.max
 import kotlin.math.min
 import kotlin.math.pow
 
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
```diff
@@ -591,13 +592,43 @@ internal class TradeInputCalculator(
...
             MarginMode.Cross -> Numeric.double.ZERO
-            MarginMode.Isolated -> if (isReduceOnly) Numeric.double.ZERO else (entryPrice - (oraclePrice ?: entryPrice)).abs() * size
+            MarginMode.Isolated -> if (isReduceOnly) {
+                Numeric.double.ZERO
+            } else {
+                val diff = when (tradeSide) {
+                    OrderSide.Buy -> entryPrice - (oraclePrice ?: entryPrice)
+                    OrderSide.Sell -> (oraclePrice ?: entryPrice) - entryPrice
+                }
+                val pnlImpact = (diff * balance * tradeLeverage) / (entryPrice + diff * tradeLeverage)
```
ruixhuang, line 629: Do we need to check for dividing by zero?

ruixhuang (approved):

src/commonMain/kotlin/exchange.dydx.abacus/calculator/TradeInputCalculator.kt
moo-onthelawn: haha i alwayyyys forget this huh

ruixhuang (approved):

PR #669, opened 2024-09-20
CLI-677 Fix issue with position list when static typing is turned on

ruixhuang:
The position list was not updated when a new position is created.  It's because that the calculated size not available when position filter is applied.


prashanDYDX (approved):
add a test for this? :)

PR #672, opened 2024-09-24
Add api state params to tracking

aforaleka:
research found out that these fields are missing
think we didn't add them back when refactoring to the different supervisors?

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/NetworkHelper.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/NetworkHelper.kt
@@ -25,6 +25,7 @@ import exchange.dydx.abacus.utils.JsonEncoder
 import exchange.dydx.abacus.utils.ParsingHelper
 import exchange.dydx.abacus.utils.ServerTime
 import exchange.dydx.abacus.utils.UIImplementations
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/NetworkHelper.kt
```diff
@@ -555,4 +556,16 @@ class NetworkHelper(
             }
         }
     }
+
+    internal fun apiStateParams(): IMap<String, Any>? {
```
ruixhuang, line 560: Nit: return type can be non-optional, e.g,. `IMap<String, Any>`

ruixhuang (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/NetworkHelper.kt
aforaleka: mm yea our ParsingHelper merge functions take in that specific type... not sure what's the best way to cast/not cast t_t

aforaleka: will send it for nowwww!

PR #680, opened 2024-09-25
correct vault validator types

tyleroooo:


ruixhuang (approved):

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultAccount.kt
```diff
@@ -68,16 +79,16 @@ object VaultAccountCalculator {
         vaultInfo: AccountVaultResponse,
         vaultTransfers: IndexerTransferBetweenResponse
     ): VaultAccount {
-        val presentValue = vaultInfo.equity
+        val presentValue = vaultInfo.equity?.let { it / 1_000_000 }
```
yogurtandjam, line 82: do we have a standard value const for `1_000_000`?

yogurtandjam (approved):

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultAccount.kt
tyleroooo: I don't think so, I'll add one in a follow up

PR #683, opened 2024-09-26
fix: pnl sorting

tyleroooo:


yogurtandjam (approved):
TEST THIS at some point!

PR #684, opened 2024-09-26
Vault share unlocks metadata

tyleroooo:
Frontends will have to calculate the predicted time: NOW + (unlock block height - latest block height) * 1s

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultAccount.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultAccount.kt
@@ -9,6 +9,7 @@ import indexer.codegen.IndexerTransferType.TRANSFER_IN
 import indexer.codegen.IndexerTransferType.TRANSFER_OUT
 import indexer.codegen.IndexerTransferType.WITHDRAWAL
 import kollections.toIList
+import kotlinx.datetime.Clock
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/VaultAccount.kt
```diff
@@ -77,14 +85,16 @@ object VaultAccountCalculator {
...
-        vaultTransfers: IndexerTransferBetweenResponse
+        vaultTransfers: IndexerTransferBetweenResponse,
     ): VaultAccount {
         val presentValue = vaultInfo.equity?.let { it / 1_000_000 }
         val netTransfers = parser.asDouble(vaultTransfers.totalNetTransfers)
         val withdrawable = vaultInfo.withdrawableEquity?.let { it / 1_000_000 }
         val allTimeReturn =
             if (presentValue != null && netTransfers != null) (presentValue - netTransfers) else null
 
+        val impliedShareValue: Double = if (vaultInfo.shares?.numShares != null && presentValue != null) presentValue / vaultInfo.shares.numShares else 0.0
```
aforaleka, line 96: can we also check if `vaultInfo.shares?.numShares` is not zero to prevent div by zero?

ruixhuang (approved):

PR #685, opened 2024-09-26
Rename validator data models for Vaults and move them to the correct package

ruixhuang:
Currently all the validator response data models are stored at package indexer.models.chain, so it's better to move the new ones there to be consistent....

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -52,7 +52,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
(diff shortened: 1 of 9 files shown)


tyleroooo (approved):
good call, though it is strange to see validator models in /indexer/models

ruixhuang:
> good call, though it is strange to see validator models in /indexer/models

Oh yeah... we should rename it to something like /backend/models at some point...

PR #693, opened 2024-10-01
fix: correct vault apr calculation

tyleroooo:
Should be 30d return but extrapolated to 1 year


src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/Vault.kt
```diff
@@ -135,7 +140,7 @@ object VaultCalculator {
 
         return VaultDetails(
             totalValue = totalValue,
-            thirtyDayReturnPercent = thirtyDayReturnPercent,
+            thirtyDayReturnPercent = thirtyDayReturnPercent * 365.days.inWholeMilliseconds / timeDifferenceMs,
```
ruixhuang, line 143: Need to check timeDifferenceMs is zero.

ruixhuang (approved):

ruixhuang (approved):

PR #710, opened 2024-10-15
chore(metadata-service): Fetch Asset information from MetadataService

jaredvu:
Add new AssetMetadataProcessor and AppConfig flag

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/configs/V4StateManagerConfigs.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/manager/configs/V4StateManagerConfigs.kt
@@ -55,7 +55,8 @@ class V4StateManagerConfigs(
                          "status":"/v1/status"
                       },
                       "configs":{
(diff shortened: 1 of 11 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/manager/configs/V4StateManagerConfigs.kt
```diff
@@ -55,7 +55,8 @@ class V4StateManagerConfigs(
                          "status":"/v1/status"
                       },
                       "configs":{
-                         "markets":"/configs/markets.json"
+                         "markets":"/configs/markets.json",
+                         "assets": "https://66iv2m87ol.execute-api.ap-northeast-1.amazonaws.com/mainnet/metadata-service/v1/info"
```
ruixhuang, line 59: Do you think it's better to put this into env.json?  In case the host ever changes, we can just update the config json.  Assets are essential to the app.

ruixhuang (approved):

PR #712, opened 2024-10-15
feat: unallocated usdc in vault positions table

tyleroooo:


mike-dydx (approved):

src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/Vault.kt
```diff
@@ -188,7 +212,29 @@ object VaultCalculator {
                 perpetualMarket = market,
             )
         }
-        return VaultPositions(positions = positions?.toIList())
+
+        if (vault.details?.totalValue != null) {
+            val usdcTotal = vault.details.totalValue - positions.sumOf { it.marginUsdc ?: 0.0 }
+
+            // add a usdc row
+            positions = positions + VaultPosition(
```
ruixhuang, line 220: Nit:  Duplication of line 172, so probably better to separate this out as a function.

tyleroooo: lol it's all duplicated though :) 

(adding a helper function now)

ruixhuang (approved):

PR #715, opened 2024-10-17
chore(metadata-url): Update assetId and get metadataService from env.json

jaredvu:
- `assetId` was previously stripped of `chain` and `address`. Restore full `assetId`
- Add `metadataServiceInfo()` to utilize `env.json` from web deployment


src/commonMain/kotlin/exchange.dydx.abacus/output/Asset.kt
```diff
@@ -112,3 +112,16 @@ data class Asset(
         }
     }
 }
+
+@JsExport
+@Serializable
+enum class AssetTags(val rawValue: String) {
```
ruixhuang, line 118: Where is this getting used?

jaredvu: I was thinking that they could be exported for use purely on the client side

jaredvu: the raw values are the coinmarketcap sector_tags that we want to support

ruixhuang (approved):

PR #724, opened 2024-10-22
Set route gasPrice and gasLimit to null

ruixhuang:
We used to get gas values for the transfer route with Squid and pass them along to the wallet.  But with Skip we no longer get them.   It's better to set them to null for the wallet to figure out the values.

### DIFF
--- a/v4_abacus.podspec
+++ b/v4_abacus.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'v4_abacus'
-    spec.version                  = '1.13.6'
+    spec.version                  = '1.13.7'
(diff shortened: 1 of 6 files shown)


yogurtandjam (approved):

v4_abacus.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'v4_abacus'
-    spec.version                  = '1.13.6'
+    spec.version                  = '1.13.7'
```
yogurtandjam, line 3: since we're removing properties from state this is technically a breaking change, right?

ruixhuang: No, it doesn't change the state exposed to the client.  We only removed it from the internal state.

PR #727, opened 2024-10-23
update calculate pnl function to use `IList` instead of `List`

mike-dydx:


tyleroooo (approved):
WHOO BREAKING CHANGES LETS GOO

PR #729, opened 2024-10-24
Fix Skip transfer status

ruixhuang:
The hash coming back from the skip endpoint doesn't have "0x" and is uppercased...

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipTrackProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipTrackProcessor.kt
@@ -12,8 +12,11 @@ internal class SkipTrackProcessor(
         payload: Map<String, Any>
     ): Map<String, Any>? {
         val modified = existing?.mutable() ?: mutableMapOf()
-        val txHash = parser.asString(payload.get("tx_hash")) ?: return modified
-        modified[txHash] = true
+        var txHash = parser.asString(payload.get("tx_hash")) ?: return modified
+        if (!txHash.startsWith("0x")) {
+            txHash = "0x$txHash"
+        }
+        modified[txHash.lowercase()] = true
         return modified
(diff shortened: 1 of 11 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/OnboardingSupervisor.kt
```diff
@@ -855,6 +856,10 @@ internal class OnboardingSupervisor(
         helper.post(url, null, body.toJsonPrettyPrint()) { _, response, httpCode, _ ->
             if (response != null) {
                 update(stateMachine.routerTrack(response), oldState)
+                val isTracked = oldState?.trackStatuses?.get(hash) == true
+                if (isTracked) {
+                    fetchTransferStatusSkip(hash, fromChainId)
```
ruixhuang, line 861: Request status right away instead of waiting for the next timer trigger.

mike-dydx (approved):

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipTrackProcessor.kt
```diff
@@ -12,8 +12,11 @@ internal class SkipTrackProcessor(
         payload: Map<String, Any>
     ): Map<String, Any>? {
         val modified = existing?.mutable() ?: mutableMapOf()
-        val txHash = parser.asString(payload.get("tx_hash")) ?: return modified
-        modified[txHash] = true
+        var txHash = parser.asString(payload.get("tx_hash")) ?: return modified
+        if (!txHash.startsWith("0x")) {
+            txHash = "0x$txHash"
```
yogurtandjam, line 17: i'm not sure we always want this though, what if the user submits a non EVM transaction? the tx hash may not necessarily start with 0x, right? 
@ruixhuang

PR #732, opened 2024-10-24
Reduce the unnecessary network reconnect

ruixhuang:
During the iOS app starts up, the  _environment is set twice with the same value.  This causes the indexer websocket to reject the second try due to rate-limiting.  We should only do reconnect when there is a real change (ie. env id is different).


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
```diff
@@ -89,8 +89,8 @@ class AsyncAbacusStateManagerV2(
 
     private var environments: IList<V4Environment> = iListOf()
         set(value) {
-            field = value
-            ioImplementations.threading?.async(ThreadingType.abacus) {
```
ruixhuang, line 93: No need for threading switch 'cos it's called from the same thread.

mike-dydx (approved):

PR #733, opened 2024-10-25
Fix transfer asset loading issues on iOS with static typing

ruixhuang:
When iOS passes in transfer.type == nil, Abacus should still set the inputType to "transfer" so that the transfer assets get assigned to the transfer.input.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+TransferInput.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+TransferInput.kt
@@ -44,11 +44,6 @@ fun TradingStateMachine.transfer(
     environment: V4Environment,
 ): StateResponse {
     if (staticTyping) {
-        if (type == null) {
-            val error = ParsingError.cannotModify("type")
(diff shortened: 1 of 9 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+TransferInput.kt
```diff
@@ -44,11 +44,6 @@ fun TradingStateMachine.transfer(
     environment: V4Environment,
 ): StateResponse {
     if (staticTyping) {
-        if (type == null) {
-            val error = ParsingError.cannotModify("type")
-            return StateResponse(state, null, iListOf(error))
-        }
```
jaredvu, line 50: are we removing because because type can be `null`?

jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/model/TradingStateMachine+TransferInput.kt
ruixhuang: Yeah.. iOS is passing null there

PR #734, opened 2024-10-25
Add parsing logic for new markets in v4_markets channel update.

ruixhuang:
This is to support user submitted new permissionless market.


jaredvu (approved):
🔥

PR #735, opened 2024-10-29
Fix a crash on iOS release build

ruixhuang:
The release build performs some code optimization that crashes the app.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/utils/TradeValidationTracker.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/utils/TradeValidationTracker.kt
@@ -3,6 +3,7 @@ package exchange.dydx.abacus.utils
 import exchange.dydx.abacus.protocols.AnalyticsEvent
 import exchange.dydx.abacus.protocols.TrackingProtocol
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/utils/TradeValidationTracker.kt
```diff
@@ -14,7 +15,8 @@ class TradeValidationTracker(
     fun logValidationResult(payload: TradeValidationPayload) {
         if (payload == lastSeen) return
 
-        trackingProtocol?.log(AnalyticsEvent.TradeValidation.name, TrackingJson.encodeToString(TradeValidationPayload.serializer(), payload))
```
jaredvu, line 17: what was the issue with the `serializer`?

jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/utils/TradeValidationTracker.kt
ruixhuang: No.. it's the in-line call.  Breaking it down to two lines fixed the problem.... It's definite a bug with code optimization.

PR #743, opened 2024-11-11
Static typing: Fix code_gen of enum values with underscores

ruixhuang:
The codegen tool removes the underscores of the enums, which causes serialization issue.  This PR fixes the issue.


tyleroooo (approved):
okie

PR #745, opened 2024-11-12
fix: trigger cancels when markets is slow

tyleroooo:


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
```diff
@@ -291,6 +291,8 @@ internal class SubaccountTransactionSupervisor(
 
     internal fun cancelTriggerOrdersWithClosedOrFlippedPositions() {
         val subaccount = stateMachine.state?.subaccount(subaccountNumber) ?: return
+        if (stateMachine.state?.marketsSummary?.marketIds()?.isEmpty() != false) return
```
tyleroooo, line 294: meant to catch true or null

jaredvu (approved):

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
ruixhuang: Nit: There is `isNullOrEmpty()` that can make this more readable

ruixhuang (approved):

PR #750, opened 2024-11-13
fix: staking delegations static typed correctly 

tyleroooo:
Unstaking was entirely broken

### DIFF
--- a/src/commonMain/kotlin/indexer/models/chain/OnChainDelegationResponse.kt
+++ b/src/commonMain/kotlin/indexer/models/chain/OnChainDelegationResponse.kt
@@ -10,10 +10,15 @@ data class OnChainDelegationResponse(
 
 @Serializable
 data class OnChainDelegationObject(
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/indexer/models/chain/OnChainDelegationResponse.kt
```diff
@@ -10,10 +10,15 @@ data class OnChainDelegationResponse(
 
 @Serializable
 data class OnChainDelegationObject(
+    val delegation: OnChainDelegationInfo? = null,
```
ruixhuang, line 13: Hmmm. Is the change coming from codegen?

tyleroooo: no I didn't codegen, this is just what the object is

ruixhuang (approved):

PR #752, opened 2024-11-14
Vault PNL cut off

ruixhuang:


src/commonMain/kotlin/exchange.dydx.abacus/functional/vault/Vault.kt
```diff
@@ -114,7 +114,7 @@ object VaultCalculator {
             }
         }.filter { entry -> entry.date != null && entry.date >= dataCutoffMs }
 
-        val latestEntry = history.first()
```
ruixhuang, line 117: first() will throw exception when list is empty.

tyleroooo: oh very good call

src/commonTest/kotlin/exchange.dydx.abacus/tests/payloads/AbacusMockData.kt
```diff
@@ -60,6 +60,7 @@ class AbacusMockData {
         chainName = "dYdX Chain",
         chainLogo = "dYdX-logo.png",
         rewardsHistoryStartDateMs = "1704844800000",
+        megavaultHistoryStartDateMs = 1704844800000.0,
```
tyleroooo, line 63: it's a string in the env.json I think

src/commonMain/kotlin/exchange.dydx.abacus/state/manager/Environment.kt
```diff
@@ -430,6 +430,7 @@ class V4Environment(
     val chainName: String?,
     val chainLogo: String?,
     rewardsHistoryStartDateMs: String,
+    val megavaultHistoryStartDateMs: Double?,
```
tyleroooo, line 433: yeah it's a string

ruixhuang: We will parse it as double.. since that's how it's going to be used.

jaredvu (approved):

PR #756, opened 2024-11-16
fix: keep old orders around on websocket reconnect

tyleroooo:
So we can render old notifications.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
@@ -124,11 +124,14 @@ internal open class SubaccountProcessor(
         state.calculated[CalculationPeriod.current] = subaccountCalculated
 
         val fills = parser.asTypedList<IndexerCompositeFillObject>(content["fills"])
-        state = processFills(
-            subaccount = state,
-            payload = fills,
-            reset = false,
-        )
+        // fills is sometimes null on this message, doesn't mean there are none
+        if (fills != null) {
+            state = processFills(
+                subaccount = state,
+                payload = fills,
+                reset = false,
+            )
+        }
 
         val orders = parser.asTypedList<IndexerCompositeOrderObject>(content["orders"])
         state = processOrders(
@@ -250,7 +253,6 @@ internal open class SubaccountProcessor(
             existing.assetPositions = assetPositionsProcessor.process(
                 payload = payload.assetPositions,
             )
-            existing.orders = null
         }
 
         val subaccountCalculated = existing.calculated[CalculationPeriod.current] ?: InternalSubaccountCalculated()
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
```diff
@@ -124,11 +124,14 @@ internal open class SubaccountProcessor(
         state.calculated[CalculationPeriod.current] = subaccountCalculated
 
         val fills = parser.asTypedList<IndexerCompositeFillObject>(content["fills"])
-        state = processFills(
-            subaccount = state,
-            payload = fills,
-            reset = false,
-        )
+        // fills is sometimes null on this message, doesn't mean there are none
+        if (fills != null) {
```
tyleroooo, line 128: I think this is why the fills were disappearing earlier too. This message has null fills even though there are some.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
```diff
@@ -250,7 +253,6 @@ internal open class SubaccountProcessor(
             existing.assetPositions = assetPositionsProcessor.process(
                 payload = payload.assetPositions,
             )
-            existing.orders = null
```
tyleroooo, line 253: not sure why we did this, also in the non-static codepath. I get that this message contains orders so we can overwrite it, but then we miss out on filled orders in that table.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: reset = false is passed here, so it shouldn't overwrite the existing fills when payload is null.

tyleroooo: even when it's true fills is null though

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: This fix makes sense... Just need to test the scenarios of user logging out or changing the env.  I think the entire subaccount state should be nuked in those cases, so that the order list becomes empty.

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/SubaccountProcessor.kt
ruixhuang: I mean processFills() handles the case when fills == null already... if fill == null, the existing fills won't be overwritten

tyleroooo:
```
subaccount.fills = fillsProcessor.process(
            existing = if (reset) null else subaccount.fills,
            payload = payload ?: emptyList(),
            subaccountNumber = subaccount.subaccountNumber,
        )
```
        
 if reset is true it's going to nuke it, which we don't want because payload is also null

ruixhuang (approved):

tyleroooo:
Feels like the upside of merging this isn't really worth the possible downside so close to unlimited cut.

PR #759, opened 2024-11-20
fix: add 250ms delay to order placement

jaredvu:


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
```diff
@@ -652,13 +654,24 @@ internal class SubaccountTransactionSupervisor(
...
-                    transactionType = TransactionType.PlaceOrder,
-                    transactionPayloadString = string,
-                    onSubmitTransaction = onSubmitOrderTransaction,
-                    transactionCallback = orderTransactionCallback,
-                    useTransactionQueue = useTransactionQueue,
-                )
+                // Return submitTransaction after a delay to ensure the transfer is confirmed
+                val timer = helper.ioImplementations.timer ?: CoroutineTimer.instance
+
+                val delayedPlaceOrder = timer.schedule(
```
ruixhuang, line 660: "timer.run()" is easier

`val delayedPlaceOrder = timer.run(delay = 0.25) { ... }`

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
```diff
@@ -652,13 +654,24 @@ internal class SubaccountTransactionSupervisor(
...
+                        transactionType = TransactionType.PlaceOrder,
+                        transactionPayloadString = string,
+                        onSubmitTransaction = onSubmitOrderTransaction,
+                        transactionCallback = orderTransactionCallback,
+                        useTransactionQueue = useTransactionQueue,
+                    )
+                    true
+                }
+
+                delayedPlaceOrder.cancel()
```
ruixhuang, line 674: No need for "cancel()"... it would cancel the timer before it runs.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/SubaccountTransactionSupervisor.kt
ruixhuang, line 660: Oh..also better make delayedPlaceOrder an ivar of the class instead of local var.

ruixhuang (approved):

PR #761, opened 2024-12-02
Fix exception when closing a position of > 100% margin usage

ruixhuang:
Also adding tracking code to log client-side exceptions like this.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarketOrderCalculator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarketOrderCalculator.kt
@@ -109,7 +109,12 @@ internal class TradeInputMarketOrderCalculator() {
         val tradeSize = trade.size
         val freeCollateral = subaccount?.calculated?.get(CalculationPeriod.current)?.freeCollateral
 
-        if (tradeSize != null && tradeSide != null && freeCollateral != null && freeCollateral > Numeric.double.ZERO) {
+        val freeCollateralCondition = if (trade.reduceOnly) {
+            true
+        } else {
+            freeCollateral != null && freeCollateral > Numeric.double.ZERO
+        }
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/calculator/V2/TradeInput/TradeInputMarketOrderCalculator.kt
```diff
@@ -109,7 +109,12 @@ internal class TradeInputMarketOrderCalculator() {
         val tradeSize = trade.size
         val freeCollateral = subaccount?.calculated?.get(CalculationPeriod.current)?.freeCollateral
 
-        if (tradeSize != null && tradeSide != null && freeCollateral != null && freeCollateral > Numeric.double.ZERO) {
+        val freeCollateralCondition = if (trade.reduceOnly) {
```
ruixhuang, line 112: Don't need to check freeCollateral > 0 if the trade is reduceOnly.

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
```diff
@@ -506,50 +508,111 @@ class AsyncAbacusStateManagerV2(
     }
 
     override fun placeOrderPayload(): HumanReadablePlaceOrderPayload? {
-        return adaptor?.placeOrderPayload()
+        try {
+            return adaptor?.placeOrderPayload()
+        } catch (e: Exception) {
+            val error = V4TransactionErrors.error(null, e.toString())
+            trackTransactionError("placeOrderPayload", error)
+            throw e
```
ruixhuang, line 516: All those "..Payload" functions are called by the web FE.. I think web code catches the exception to display the error. 

Here we just catch the exception for logging and re-throw it.

jaredvu (approved):
Nice. I think web might also be logging these errors as well. will see if this ends up in duplicate logs

PR #764, opened 2024-12-03
Fix Polygon route

ruixhuang:
The Polygon route in cctp.json has the same value as what Skip returns but different cases.  So the comparison needs to be case-insensitive.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -52,7 +52,7 @@ allprojects {
 }
 
 group = "exchange.dydx.abacus"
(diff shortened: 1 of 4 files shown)


tyleroooo (dismissed):
weren't there comparisons in the other pr that didn't do the lowercase() checks too?

ruixhuang:
> weren't there comparisons in the other pr that didn't do the lowercase() checks too?

Yeah.. added one more... Please restamp

tyleroooo (approved):

PR #770, opened 2024-12-21
BUG2-214 Showing backend error for TP/SL orders

ruixhuang:
Fix https://linear.app/dydx/issue/BUG2-214/unable-to-add-tpsl-on-mad-market-it-gets-set-as-price-dollar0

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
@@ -34,10 +34,19 @@ class TriggerOrderToastGenerator(
     private val localizer: LocalizerProtocol?,
     private val threading: ThreadingProtocol?,
 ) : TriggerOrderToastGeneratorProtocol {
-    private enum class TriggerOrderStatus {
-        Submitting,
-        Success,
-        Failed
+    private sealed class TriggerOrderStatus {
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
```diff
@@ -348,17 +371,21 @@ class TriggerOrderToastGenerator(
...
-        val detail = detailStringPath?.let {
+        var detail = detailStringPath?.let {
             localizer.localizeWithParams(
                 path = it,
                 params = mapOf("NEW_VALUE" to triggerPrice),
             )
         }
+        val failedReasons = status.failedReason
+        if (failedReasons != null) {
+            detail += " $failedReasons"
```
jaredvu, line 382: Does this coming the localized error message with the broadcast error?

ruixhuang:
Yup.  Web does its own message and this code path is for mobile only

src/commonMain/kotlin/exchange.dydx.abacus/state/app/helper/TriggerOrderToastGenerator.kt
ruixhuang: Yup.  Web does its own message and this code path is for mobile only

tinaszheng (approved):

PR #778, opened 2025-02-04
Fix isolated position withdrawal issue when there is no equity

ruixhuang:
Use childSubaccountErrors for trade/close-position on isolated positions.
Also updating staticTyping for tests.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/output/input/Input.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/output/input/Input.kt
@@ -35,7 +35,6 @@ data class Input(
     val adjustIsolatedMargin: AdjustIsolatedMarginInput?,
     val receiptLines: IList<ReceiptLine>?,
     val errors: IList<ValidationError>?,
-    val childSubaccountErrors: IList<ValidationError>?
 ) {
     companion object {
         internal fun create(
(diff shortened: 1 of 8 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/output/input/Input.kt
```diff
@@ -104,22 +103,12 @@ data class Input(
                     )
                 }
 
-                val errors = if (staticTyping) {
-                    internalState?.input?.errors?.toIList()
-                } else {
-                    ValidationError.create(existing?.errors, parser, parser.asList(data?.get("errors")))
+                var errors = internalState?.input?.errors?.toIList()
+                if (internalState?.input?.currentType == InputType.TRADE && internalState.input.trade.marginMode == MarginMode.Isolated) {
+                    errors = internalState.input.childSubaccountErrors?.toIList()
```
tyleroooo, line 108: will this just be the relevant subaccount? could errors from irrelevant subaccounts fall in? Is this excluding errors from the primary subaccount in cases where the trade will require a transfer from the primary to a secondary subaccount?

ruixhuang: The validator runs on the subaccount where the current input happens, so yeah there won't be other subaccounts mixed in.

I don't know the answer to the second question.  @jaredvu wrote the logic, so maybe he remembers.  Otherwise, we'd just land this and test it out.

jaredvu: There is always a transfer from the parentSubaccount, so I don't believe that errors from irrelevant subaccounts will sneak in

jaredvu: Sorry, my previous comment was responding to Tyler's question with the previous implementation w/o looking at the code change. @ruixhuang were you able to test this? I think that the childSubaccount might actually return freeCollateral errors. The reason we can ignore this by just using parentSubaccountErrors is because a transfer is always expected to accompany the trade.

tyleroooo (approved):

PR #780, opened 2025-02-07
Fix a race condition on tradeInMarket

ruixhuang:


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/StateManagerAdaptorV2.kt
```diff
@@ -256,14 +256,12 @@ internal class StateManagerAdaptorV2(
                 markets.marketId = value
 
                 if (value != null) {
-                    networkHelper.ioImplementations.threading?.async(ThreadingType.abacus) {
```
ruixhuang, line 259: No need to do another async threading call since the [caller](https://github.com/dydxprotocol/v4-abacus/blob/31c5283f501fdfa93e7e16380722e2b30e69dbd1/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt#L168) is already on the abacus thread.  Otherwise, there can be potential race condition when a "closePosition" is invoked right after setting the market

jaredvu (approved):

PR #782, opened 2025-02-14
Add autoStart option

ruixhuang:
The current behavior is to start the connections once AbacusStateManager is initialized.  This now causes issues on Android because the JS runtime environment is no longer initialized by the time AbacusStateManager starts calling the JS functions.

This PR adds an autoStart option (default to true).  With this option, Android can set it to false, and call "start()" only after its JS runtime finishes loading.


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/supervisor/Configs.kt
```diff
@@ -216,6 +216,7 @@ data class AppConfigsV2(
     var enableLogger: Boolean = false,
     var triggerOrderToast: Boolean = false,
     var staticTyping: Boolean = true,
+    var autoStart: Boolean = true,
```
tyleroooo, line 219: can this be false by default so we are certain the web doesn't break?

ruixhuang:
autoStart=true is the current behavior, and Android needs to set it to
false to fix its issue.

tyleroooo:
ooo I misread the description then

tyleroooo (approved):

PR #789, opened 2025-03-10
Fix a bug in Android startup

ruixhuang:
Android needs to delay the initialization sequence until after its webview is loaded.  I added this logic from a previous PR, but missed that there is another "init" block in the AsyncAbacusStateManagerV2 class (apparently Kotlin is okay with multiple init blocks in a class).  This causes sporadic trade failure on Android due to the race condition.

### DIFF
--- a/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
+++ b/src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
@@ -91,12 +91,16 @@ class AsyncAbacusStateManagerV2(
             Logger.isDebugEnabled = true
         }
         if (appConfigs.autoStart) {
+            initAfterStart()
             started = true
         }
     }
 
     override fun start() {
-        started = true
+        if (!started) {
+            initAfterStart()
+            started = true
+        }
     }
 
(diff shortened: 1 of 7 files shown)


src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
```diff
@@ -91,12 +91,16 @@ class AsyncAbacusStateManagerV2(
             Logger.isDebugEnabled = true
         }
         if (appConfigs.autoStart) {
+            initAfterStart()
```
ruixhuang, line 94: iOS and web go through this codepath

src/commonMain/kotlin/exchange.dydx.abacus/state/v2/manager/AsyncAbacusStateManagerV2.kt
```diff
@@ -91,12 +91,16 @@ class AsyncAbacusStateManagerV2(
...
             started = true
         }
     }
 
     override fun start() {
-        started = true
+        if (!started) {
```
ruixhuang, line 100: Android calls this after its webview is loaded

tyleroooo (approved):

tyleroooo (approved):

PR #797, opened 2025-03-25
Fix transfer status logic to support Solana withdrawal

ruixhuang:
See comments in the code.

The rest of the changes are just code clean-up.


src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipTrackProcessor.kt
```diff
@@ -5,25 +5,28 @@ import exchange.dydx.abacus.protocols.ParserProtocol
...
+            }
+            if (txHash.uppercase() == workingHash.uppercase()) {
+                modified[hash] = true
+            }
         }
-
-        if (!txHash.startsWith("0x")) {
-            txHash = "0x$txHash"
-        }
-        modified[txHash.lowercase()] = true
```
ruixhuang, line 26: The txHash returned from Skip might be of difference case, so we shouldn't use it.  Instead, the status should be keyed by the hash value sent to the Skip status endpoint.

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/IRouterProcessor.kt
jaredvu, line 1: 🙏

src/commonMain/kotlin/exchange.dydx.abacus/processor/router/skip/SkipTrackProcessor.kt
```diff
@@ -5,25 +5,28 @@ import exchange.dydx.abacus.protocols.ParserProtocol
...
-        if (explorerLink != null && explorerLink.contains("solscan.io")) {
-            // Solana tx hashes are case-sensitive
-            modified[txHash] = true
-            return modified
+        var txHash = parser.asString(payload["tx_hash"])
+        if (txHash != null) {
+            // txHash could be returned with or without the 0x prefix, and cases can be different
+            if (txHash.startsWith("0x")) {
+                txHash = txHash.substring(2)
+            }
```
jaredvu, line 21: why strip the 0x at all? Doesn't the tracking endpoint require it?

jaredvu: oh I understand the comment now

jaredvu (approved):

PR #816, opened 2025-06-17
Add funding payments processing

ruixhuang:


src/commonMain/kotlin/exchange.dydx.abacus/output/account/SubaccountFundingPayment.kt
```diff
@@ -17,66 +15,21 @@ data class SubaccountFundingPayment(
...
-                    parser.asMap(itemData),
-                )
-            })?.toIList()
+            new: IList<SubaccountFundingPayment>?,
+        ): IList<SubaccountFundingPayment> {
+            return ParsingHelper.merge(
+                existing = existing,
+                new = new,
+                comparison = { obj, newItem ->
+                    ParsingHelper.compare(obj.effectiveAtMilliSeconds, newItem.effectiveAtMilliSeconds, true)
```
jaredvu, line 29: I don't think `effectiveAtMilliSeconds` exist in the Indexer response. Are you creating this key later in the pipeline?

src/commonMain/kotlin/exchange.dydx.abacus/processor/wallet/account/FundingPaymentsProcessor.kt
```diff
@@ -0,0 +1,35 @@
...
+                existing = null,
+                payload = eachPayload,
+            )
+        }
+        return merge(
+            parser = parser,
+            existing = existing,
+            incoming = new,
+            timeField = { item ->
+                item?.effectiveAtMilliSeconds?.toLong()?.let {
```
jaredvu, line 28: same question here. Also I think if you keep this key

`effectiveAtMilliSeconds` -> `effectiveAtMilliseconds`

ruixhuang: effectiveAtMilliSeconds is part of the output structure, so it should be fine...

jaredvu (dismissed):

